PR #1, opened 2017-01-18
Drone build for web and API

jits:
Runs tests, builds docker images, tests the docker images, and pushes the docker images to Quay on a `master` branch push.


vaijab:
Looks sensible.

platform-hub-api/Dockerfile
```diff
@@ -0,0 +1,18 @@
...
+
+WORKDIR /app
+
+ENV RAILS_ENV production
+ENV RACK_ENV production
+
+COPY Gemfile Gemfile.lock ./
+RUN bundle install --jobs 20 --retry 5 --deployment --without development test --no-cache
+
+COPY . .
```
marcinc: I guess you could just `bundle` after that and skip two lines above?

jits: Discussed offline – it's done this way to benefit from the Docker layer caching :)

PR #2, opened 2017-01-24
Backend persistence of User and Identity records + GitHub identity flows 

jits:
See the individual commits for details.


vaijab:
It's difficult for me to review these changes since I don't know rails, nor javascript that much.

platform-hub-web/src/app/home/home.module.js
```diff
@@ -2,7 +2,7 @@ import angular from 'angular';
 
 import {HomeComponent} from './home.component';
 
-// Define the app section name
+// Main section component name
 export const AppHome = 'appHome';
```
Joseph-Irving: This is probably just my lack of angular knowledge but I'm not 100% sure what this file is for. You'll have to teach me this bit

jits: Hey @Joseph-Irving – I'm roughly following these guidelines / styleguide: https://github.com/toddmotto/angular-styleguide – it advocates splitting your app up into fairly isolated and decoupled modules (in AngularJS a module is just a grouping of injectables services/factories/components/etc.). The pattern that this particular file follows is: every top level section of the app should have it's own module, and that module is then made a dependency of the main app. That way all specific code for that section goes into it's relevant module, but app-wide shared code goes into the `app.shared` module.

jits:
@marcinc –

> Does it make sense to document API as we go with raml or swagger?

At some point, yeah, probably :) Perhaps post-MVP though ;-)

platform-hub-api/app/controllers/concerns/authentication.rb
```diff
@@ -0,0 +1,50 @@
...
+    end
+
+    return nil if payload.blank? || payload['sub'].blank?
+
+    id = payload['sub']
+
+    user = User.find_by_id(id)
+
+    if user.blank?
+      user = NewUserService.new.create payload
```
jits: I'm not a fan of static methods since they become one instance for the _whole_ runtime, instead of isolating effects to as small an area as possible (albeit this particular service is stateless and behaves in a highly functional way). Given this is happening within a request/response cycle, it's fairly common to have service objects be instantiated and used just the cycle.

In the future, if this becomes a perf problem, we can have a higher level area in which to instantiate singleton-esque instances of various services and then inject them we where need (makes it nicer to test).

platform-hub-api/app/services/new_user_service.rb
```diff
@@ -0,0 +1,28 @@
...
+      external_id: auth_token_payload['sub'],
+      external_username: auth_token_payload['preferred_username'],
+      external_name: auth_token_payload['name'],
+      external_email: auth_token_payload['email'],
+      data: auth_token_payload
+    )
+
+    user.last_seen_at = DateTime.now
+
+    user.save!
```
jits: In theory:

1. All/most validation should be handled by the model.
1. The data coming from the token is expected to be trusted as we expect the auth proxy to handle all validation of the token.

Having said that, can you think of any extra validations we can/should add here?

marcinc: Basically you save with a bang which means if there is any issue with validation it'll blow.

jits: Yep – that's what I want :) Hoping that it's better to fail early and catch in dev/test. (Note: given it's an API endpoint, the frontend app will catch the failed API request and show and error message to the user).

platform-hub-api/app/controllers/concerns/authentication.rb
marcinc: > (albeit this particular service is stateless and behaves in a highly functional way) 

+1 for functional. At the moment the only thing you want from this services is to create a new user and the service itself is stateless so no point in creating an instance of it every time. I'm not fussed at this stage tho. Can be easily changed moving forward.

platform-hub-api/app/services/new_user_service.rb
marcinc: 👍

platform-hub-auth-proxy/local/.env
```diff
@@ -9,3 +9,5 @@ export PHUB_AUTHPROXY_DATA_LOCAL
 export PHUB_AUTHPROXY_DATA_CONTAINER="/opt/data"
 
 export PHUB_AUTHPROXY_CONTAINER_NAME=phub-auth-proxy
+
+export KCPROXY_UPSTREAM_URL="http://10.200.10.1:8081"
```
danielepolencic: with a docker network/docker compose file this should be "http://<name-of-keycloack-service>:8081". No IP hardcoded

platform-hub-web/src/app/identities/git_hub_identity.service.js
```diff
@@ -0,0 +1,18 @@
+export const gitHubIdentityService = function (windowPopupService, apiEndpoint, homeEndpoint) {
+  'ngInject';
+
+  const service = {};
+
+  service.popupFlow = popupFlow;
```
danielepolencic: this two lines can be condensed to:

```js
const service = {popupFlow};
```

platform-hub-web/src/app/identities/identities-manager.component.js
```diff
@@ -0,0 +1,54 @@
...
+  'ngInject';
+
+  const ctrl = this;
+
+  ctrl.supported = [
+    {provider: 'keycloak', title: 'Keycloak (using Office 365)', external: false},
+    {provider: 'github', title: 'GitHub', external: true}
+  ];
+
+  ctrl.identities = {};
```
danielepolencic: same as above, consider initialising the value within the object: 

```js
const ctrl = {identities: ..., connect: ...};
```

platform-hub-web/src/app/identities/identities-manager.component.js
```diff
@@ -0,0 +1,54 @@
...
+  function connect(service) {
+    // Currently only support Github
+    if (service !== 'github') {
+      return;
+    }
+
+    gitHubIdentityService
+      .popupFlow()
+      .then(hubApiService.getMe);
+  }
```
danielepolencic: the above function has signature: `Promise<any> or undefined`. Consider returning a promise all the time:

```js
if (service !== 'github') {
      return Promise.reject('Error');
    }
```

platform-hub-api/Dockerfile
```diff
@@ -1,6 +1,6 @@
 FROM ruby:2.3.3
```
jits: We do, but it makes too many assumptions (via it's `ONBUILD` steps) so it's unusable for this case. We should do a "plainer" image at some point, but just to unlock this work I'm using the official Ruby image. We can certainly revisit this before we deploy to prod.

platform-hub-api/app/services/git_hub_api_service.rb
```diff
@@ -0,0 +1,19 @@
...
+
+  def authorize_url client_id:, scope:, redirect_uri:, state:
+    @client.authorize_url(client_id, scope: scope, redirect_uri: redirect_uri, state: state)
+  end
+
+  def user
+    @client.user
+  end
+
+end
```
jits: Fair point. Initially I was expecting this to have more "wrapped up" functionality, but it didn't in the end. In any case these wrappers are really useful for mocking out in tests for the consumers. Otherwise I'd have to either a) dependency inject the `Octokit` class, or b) mock out the global class itself in a test. Mind you, it's not a big difference either way!

Also, I notice now that I'm using `Octokit` class directly in the consumer anyway. We should make this consistent either way, so I'll go with your suggestion, until we have a need for a separate wrapper 👍

platform-hub-api/config/secrets.yml
```diff
@@ -10,13 +10,17 @@
...
+  github_client_id: <%= ENV.fetch('GITHUB_CLIENT_ID') { raise 'GITHUB_CLIENT_ID missing from env' } %>
+  github_client_secret: <%= ENV.fetch('GITHUB_CLIENT_SECRET') { raise 'GITHUB_CLIENT_SECRET missing from env' } %>
+
+
 development:
-  secret_key_base: c6c7fb336cd9983d677318ccca486149658bb14eb9f88a0fddf90f936d4e5d328ffc5d1c06576f99b4df6659e45e216f1917abbc9c17a186e252fce424c2653e
+  <<: *common
 
 test:
-  secret_key_base: afd76d611a7ae59d56ead6888e6b6b0f51f31e1fbbf4e865eb3f19e0bbfde9e9e7c1f5e8d1dc2888888c7d41417a4310f221749384a46fe6b9bd672f491bccfa
```
jits: Not this key specifically, but the one provided via an env var. It's used to encrypt the ID of the user account during the OAuth flow with GitHub (in the `state` parameter), so we can know – in the callback – which user to associate that GitHub identity with.

platform-hub-auth-proxy/README.md
```diff
@@ -4,7 +4,11 @@ Uses [keycloak-proxy](https://github.com/gambol99/keycloak-proxy)
 
 The `local` folder contains scripts to manage a local `keycloak-proxy` instance in a Docker container.
 
+**IMPORTANT:** to allow the Docker container to talk to the API running on your host: set up an alias on your local loopback adapter using `sudo ifconfig lo0 alias 10.200.10.1/24`.
```
jits: Yeah, I'm not 100% happy with this solution TBH, but it's intended to be a quick and mostly viable solution for local dev.

I'm not keen on a fully docker compose or minikube setup as I've found that things like auto reload of code, etc. don't work so well when developing on your machine but running the services in a containerised or VM environment. Right now I'm only running things in containers that are essentially just external dependencies (Postgres and the keycloak-proxy).

Will quiz you on setting a local Docker network, at some point :)

platform-hub-auth-proxy/local/.env
jits: Sounds dreamy :) Will pick your brains on this, soon.

platform-hub-web/src/app/identities/git_hub_identity.service.js
jits: I'm following a pattern here for service definitions: even if it just the one method exported it's (slightly) nicer when adding new methods in the future to declare things like this (not to mention consistent when glossing over files).

platform-hub-web/src/app/identities/identities-manager.component.js
jits: I can't do that in this particular case because `ctrl` is just a reference to `this` and I want to bind these things to the context provided by `this`.

platform-hub-web/src/app/identities/identities-manager.component.js
jits: I don't believe this function is returning anything?

danielepolencic: You're right - ignore the above.

platform-hub-web/src/app/shared/auth/auth.service.js
```diff
@@ -99,16 +90,7 @@ export const authService = function ($window, $cookies, $q, $filter, $base64, $s
...
-      width: popupHeight,
-      height: popupWidth,
-      top: $window.screenY + (($window.outerHeight - popupHeight) / 2.5),
-      left: $window.screenX + (($window.outerWidth - popupWidth) / 2)
-    };
-  }
-
   function broadcastAuthData() {
-    $rootScope.$broadcast(authDataEvent, getPayload());
+    $rootScope.$broadcast(events.auth.updated, getPayload());
```
danielepolencic: Using `$broadcast` and `$on` is generally not a good practise, since you're dealing with an object that is passed around a global. Since this is a PoC is probably fine, but you may want to refactor this into a model.

jits: Agreed it's not good practice _in general_ – when overused and when being too liberal about what data updates get shared using such global events.

In this particular case, the two events I've defined are _intended_ to share the two global objects across the app (auth status and "me" profile data) so that any component that relies on them can react accordingly. I'd certainly question it's use beyond these two cases (since any further data updates should most likely be isolated to a particular section / component, and maybe it's children). At that point I'd start using something like [ng-redux](https://github.com/angular-redux/ng-redux).

In terms of these data objects being global, if any component/service wants to derive data or augment them then they should clone them first (this is a general rule for any part of the app really, to keep things a bit more functional).

How would refactoring it into a model help in this case? (Note that I need to be able to listen out for changes from various different places in the app, and would rather not use `$scope.$watch(() => { monitor_service_method_for_changes }, () => { … })`).

PR #7, opened 2017-02-03
Fixes for deployments

jits:


platform-hub-api/config/boot.rb
```diff
@@ -5,16 +5,17 @@
...
 # Based on: http://stackoverflow.com/a/33852354/238287
-# See README for why we force bind to 0.0.0.0
+# See README for why we force bind to 0.0.0.0 in development mode
 
 require 'rails/commands/server'
 
-module RailsServerDefaultOptions
-  def default_options
-    super.merge!(Host: '0.0.0.0')
+if Rails.env == 'development'
```
marcinc: maybe nicer like so: `Rails.env.development?`

jits: ah yeah that's much better :)

PR #11, opened 2017-02-10
API: audit key actions performed

jits:


platform-hub-api/app/models/audit.rb
```diff
@@ -0,0 +1,67 @@
...
+  def message
+    m = []
+    m << "[Audit] Action: #{self.action}"
+    m << "By user: #{self.user.id} - #{self.user_name} - #{self.user_email}" if self.user
+    m << "On thing: #{self.auditable_type} - #{self.auditable_id} - #{self.auditable_descriptor}" if self.auditable
+    m << "Associated to: #{self.associated_type} - #{self.associated_id} - #{self.associated_descriptor}" if self.associated
+    m << "Comment: #{self.comment}" if self.comment
+    m << "Remote IP: #{self.remote_ip}" if self.remote_ip
+    m << "Request UUID: #{self.request_uuid}" if self.request_uuid
+    m.join(' | ')
```
marcinc: Would be better to produce JSON instead I think.

platform-hub-api/app/models/user.rb
```diff
@@ -21,6 +25,8 @@ class User < ApplicationRecord
     :against => :name,
     :using => :trigram
 
+  has_many :audits
```
marcinc: `audited` above would already do this?

platform-hub-api/spec/integration/auditing_integration_spec.rb
```diff
@@ -0,0 +1,50 @@
+require 'rails_helper'
```
marcinc: `auditing_spec.rb` will suffice I guess.

platform-hub-api/app/services/audit_service.rb
```diff
@@ -0,0 +1,20 @@
+module AuditService
```
marcinc: Worth adding a simple spec for this service?

platform-hub-api/spec/models/audit_spec.rb
```diff
@@ -0,0 +1,4 @@
+require 'rails_helper'
+
+RSpec.describe Audit, type: :model do
```
marcinc: Should this be empty?

platform-hub-api/app/models/audit.rb
jits: Agreed… but, later :) Right now we don't have JSON output logging set up in the API stack, but when we do, we'll need to review all log messages and we can change this then. Does that sound okay to you? Or do you think we should do this now? (I'm thinking that right now it's better to have something more bespoke and descriptive like this in the logs, but later we can just output the whole `Audit` entry as a JSON object).

platform-hub-api/app/models/user.rb
jits: Oh, good catch 👍

platform-hub-api/spec/integration/auditing_integration_spec.rb
jits: I wanted these particular spec files to stand out a little, because it goes outside of the boundaries of the "standard" spec groups that rspec-rails provides (the `integration` folder is non-standard). In this particular case I wanted a place for any kind of integration spec and so figured naming the actual files with `*_integration_spec.rb` would help later.

platform-hub-api/app/services/audit_service.rb
jits: Was debating this… it seems simple enough – just a wrapper around the model, without much mapping logic, and figured the integration test would cover enough bases?

platform-hub-api/spec/models/audit_spec.rb
jits: Ah yeah, I initially started with a model unit test, but moved to the integration test instead.

jits: Will delete.

jits:
@marcinc – updated :)

platform-hub-api/app/services/audit_service.rb
marcinc: Fair enough.

PR #13, opened 2017-02-14
Admin and project 'manager' flows to onboard (and offboard) users to GitHub

jits:
- **Only** admins can make project team members a "manager" of the project.
- **Only** admins can demote team managers.
- **Only** admins can continue to add people to a project's team – managers won't have this ability.
- Admins and project managers can add an _existing_ team member of that project to a specified GitHub org + a main GitHub team within this org.
  - Note: only if that person has connected up their GitHub account (the UI will give feedback on this to the manager).
- Admins and project managers can offboard a team member from GitHub (i.e. remove them from the GitHub team AND org [at the same time]).

Note: we're currently calling the GitHub API synchronously for these operations. In the future, we may want to break this out into async / background tasks, if needed.


platform-hub-api/app/controllers/users_controller.rb
```diff
@@ -43,14 +43,51 @@ def revoke_admin
...
+        head :no_content
+      else
+        render_error "Failed to Github #{agent_action} the user - the GitHub API may be down", :service_unavailable
+      end
+    rescue Agents::GitHubAgentService::Errors::IdentityMissing
+      render_error 'User does not have a GitHub identity connected yet', :bad_request
+    end
+  end
+
+  def gitHubAgentService
```
marcinc: Maybe move to `agents` initializer? I know it's only used here currently but when we add more it'll be already in place.

platform-hub-api/app/models/ability.rb
```diff
@@ -5,5 +5,27 @@ def initialize(user)
...
+      # Find common projects between the two
+      target_user_projects = target_user.project_ids
+      user_projects = user.project_ids
+      common_projects = target_user_projects & user_projects
+
+      # Check to see if the user taking the action is a project manager of _any_
+      # common project they are both in
+      if common_projects
+        common_projects.any? do |p_id|
+          ProjectMembership.exists?(project_id: p_id, user_id: user.id, role: 'manager')
```
marcinc: Is it possible to find out by making a single query?

platform-hub-api/spec/services/agents/git_hub_agent_service_spec.rb
```diff
@@ -0,0 +1,92 @@
+require 'rails_helper'
+
+describe Agents::GitHubAgentService, type: :service do
+
+  let :token do
+    'foo_token'
+  end
```
marcinc: minor stylistic thing but I think I prefer one liners for short values e.g. `let :token { 'foo_token' }`.

platform-hub-api/app/controllers/users_controller.rb
jits: Do you mean put the logic in this method within the constructor of the agent service itself?

platform-hub-api/app/models/ability.rb
jits: Probably not without writing some ugly manual SQL :) I figured it's highly unlikely that two people will be in multiple projects, so the probability of this query being made more than once is very low :)

platform-hub-api/spec/services/agents/git_hub_agent_service_spec.rb
jits: Yeah, fair point – will change.

jits:
TODO: fix ze build!

platform-hub-api/spec/services/agents/git_hub_agent_service_spec.rb
jits: Hm, weird… my Ruby linting tells me your example is invalid (!) – this may be why I got into the habit of just using the `do … end` style of blocks. (it should be using plain old rubocop under the hood, so not sure why it's complaining).

platform-hub-api/app/controllers/users_controller.rb
marcinc: I meant maybe we should have an `agents` initializer in `config/initializers` where we keep those?

platform-hub-api/spec/services/agents/git_hub_agent_service_spec.rb
marcinc: `let(:token) { 'a value' }` should fix it :)

jits: Ha! Of course… damn parantheses :)

platform-hub-api/app/controllers/users_controller.rb
jits: Could do – though I'm not a fan of global instances of services like that (makes it harder to test; more coupled; less flexible, etc.)

jits:
@marcinc – updated

marcinc:
🚢 🇮🇹

PR #14, opened 2017-02-15
WIP! Add the BBC Accessibility Standards Checker

jits:
https://github.com/bbc/bbc-a11y

… could provide a very basic a11y checker that we could eventually plug in to the build process.


jits:
Closing this as https://github.com/bbc/bbc-a11y is not available anymore, and we probably need to rethink this anyway.

PR #15, opened 2017-02-20
Various web stack upgrades (mostly for local dev)

jits:
_See individual commits for details_

This is mainly to upgrade to the latest Webpack (as well as other local dev dependencies).


stefancocora:
lg

PR #17, opened 2017-02-20
Various README updates

jits:
Would appreciate feedback on:

- Does it make sense?
- What would you like to see in these READMEs?
- How can we make it more concise?
- Anything else you want to talk about!


marcinc:
LGTM

PR #23, opened 2017-02-27
Web: new Help centre with an initial FAQ section

jits:
FAQ also shown on homepage

### DIFF
--- a/platform-hub-web/src/app/app.module.js
+++ b/platform-hub-web/src/app/app.module.js
@@ -18,6 +18,7 @@ import 'angular-url-encode';
 import moment from 'moment';
 import lodash from 'lodash';
 
+import {HelpModule} from './help/help.module';
 import {HomeModule} from './home/home.module';
 import {IdentitiesModule} from './identities/identities.module';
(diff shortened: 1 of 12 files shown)


marcinc:
LG. Just a thought. It would be nice to maintain FAQ in a markdown file and then have it imported but I think you use some Angular magic there with linking(?) so not sure this would work.

jits:
@marcinc – 

> It would be nice to maintain FAQ in a markdown file and then have it imported but I think you use some Angular magic there with linking(?) so not sure this would work.

Yeah, it would be nicer if these were in a markdown file, or even fully content management driven from the server. I'm thinking in the future we may need a dual approach: have some static FAQs/docs in the web app (like these), and then pull in some dynamic docs from the server, maybe for a different section of the Help Centre. Not sure yet. As you say, we may need capabilities that we can't easily plug into (like the routing states), so we can figure it out if/when we need it :)

jits:
Just occurred to me that the main use case for this was to explain why some team members might not have "Onboard to GitHub" option displayed (i.e. if they haven't connected up their GitHub identity). Will add these as a separate FAQ entry.

PR #27, opened 2017-03-13
Support requests via the hub!

jits:
- Admins can design the support request form templates they want users to see and use
- Users can select a form, fill it in, and submit it
- Support requests get submitted as an issue on the GitHub repository specified in the template
- Profit

---

TODO:

- [X] Test this out with a new dev GitHub "agent" token with `repo` permissions
- [x] Finish tests
- [x] Audit the submission of requests
- [x] Add help box on requests overview page prompting the user to connect their GitHub identity (for a better user experience)
- [x] Add FAQ entries pointing to the Support Requests section
- ~~Do we need a way of seeding some initial basic templates (e.g. a fallback single text box form)?~~
  - No, we can just create these on the environment(s) we need
- [X] Do admins need to be able to define the order in which the support request options get shown to users?
  - Yes, but future task (will add to backlog); for now order alphabetically (using shortname)


jits:
- Admins can design the support request form templates they want users to see and use
- Users can select a form, fill it in, and submit it
- Support requests get submitted to the as an issue on the GitHub repository specified in the template
- Profit

jits:
For anyone interested, there are a bunch of open questions / things to do in [the description of this PR](https://github.com/UKHomeOffice/platform-hub/pull/27#issue-213836835) that could do with more feedback / thoughts.

platform-hub-api/app/models/support_request_template.rb
```diff
@@ -0,0 +1,107 @@
...
+
+  def hash_validation_errors_to_string errors
+    errors.map do |e|
+      "-- #{e[:full_path]} should be #{e[:message]}"
+    end.join(' ')
+  end
+
+  def check_form_spec
+    if self.form_spec
+      fields = self.form_spec['fields'] || self.form_spec[:fields]
```
marcinc: Maybe `fs = HashWithIndifferentAccess.new(self.form_spec)`, then always access with symbol `fields = fs[:fields]` etc?

platform-hub-api/app/models/support_request_template.rb
```diff
@@ -0,0 +1,107 @@
...
+      if fields && !fields.empty?
+        uniq_ids = fields.map{|f| f['id'] || f[:id]}.uniq
```
marcinc: this will then be `uniq_ids = fields.map(&:id).uniq`

platform-hub-api/app/models/support_request_template.rb
```diff
@@ -0,0 +1,107 @@
...
+    presence: true
+
+  validates :git_hub_issue_spec,
+    presence: true
+
+  validate :check_schemas
+  validate :check_form_spec
+
+  class << self
+    attr_accessor :schemas, :form_field_types
```
marcinc: are these ever going to be modified? If not then maybe just `attr_reader`.

platform-hub-web/src/app/shared/hub-api/hub-api.service.js
```diff
@@ -13,16 +15,23 @@ export const hubApiService = function ($rootScope, $http, $q, logger, events, ap
...
   service.removeProjectMembership = removeProjectMembership;
   service.projectSetRole = projectSetRole;
   service.projectUnsetRole = projectUnsetRole;
   service.userOnboardGitHub = userOnboardGitHub;
   service.userOffboardGitHub = userOffboardGitHub;
+  service.getSupportRequestTemplates = buildCollectionFetcher('support_request_templates');
+  service.getSupportRequestTemplate = buildResourceFetcher('support_request_templates');
+  service.createSupportRequestTemplate = buildResourceCreator('support_request_templates');
+  service.updateSupportRequestTemplate = buildResourceUpdater('support_request_templates');
+  service.deleteSupportRequestTemplate = buildResourceDeletor('support_request_templates');
```
marcinc: `buildResourceDestroyer` ? :)

platform-hub-api/app/controllers/concerns/agents_initializer.rb
```diff
@@ -0,0 +1,12 @@
+module AgentsInitializer
+  extend ActiveSupport::Concern
+
+  def git_hub_agent_service
+    @git_hub_agent_service ||= Agents::GitHubAgentService.new(
```
jits: > Should we call those `agents` or `services`?

It does _seem_ redundant to use both terms, but 'agents' are intended to be a subset of services, that access external resources/services and often have super powers (like a token with full admin access, for example). I've tried to explain this in [the API server README](https://github.com/UKHomeOffice/platform-hub/tree/master/platform-hub-api#general-architectural-points-of-interest) – let me know if you think this needs improving.

> Should this be moved to global initializer?

I prefer not to have them be available globally, and instead, controllers (or any other services) that need access to these agents should `include` this concern instead. This way the developer had to explicitly choose access to these agents.

platform-hub-api/app/controllers/support_requests_controller.rb
```diff
@@ -0,0 +1,37 @@
...
+
+    formatted_result = SupportRequestFormatterService.format template, data, current_user
+
+    repo_url = template.git_hub_repo
+    title = formatted_result.title
+    body = formatted_result.body
+
+    begin
+      url = git_hub_agent_service.create_issue repo_url, title, body
+    rescue => e
```
jits: No specific exceptions come to mind here; this is purely to catch all exceptions (except for things like ruby runtime errors) so that we can send a fallback error message to the user that some unknown error has occurred.

platform-hub-api/app/models/support_request_template.rb
jits: Good call!

platform-hub-api/app/models/support_request_template.rb
jits: Good point. Actually, I can just use string keys version all the time. I originally added this because sometimes the check is on new data, and sometimes on data from the db. New data will (usually) be set from controller params (which is an indifferent access hash anyway), but data from the db will always be string keys (very annoying really, but a quirk of the postgres adapter).

Will change to use string keys only.

platform-hub-web/src/app/shared/hub-api/hub-api.service.js
jits: Heheh – I always struggle with these terms. From an API perspective, `DELETE` is the HTTP verb.  So I figured we'd reuse that term instead.

Joseph-Irving:
For the last two points: yeah we should have some default text box.
Admins defining the order could be interesting as it could flow in a way that starts with the most obvious and simple problems and then ends up with the more complicated or totally new problems

jits:
Thanks @Joseph-Irving.

For the default form – I'm thinking this could just be one of the forms you configure? Since all the forms are content driven, it's more work to have the server add this default form to the list (and use it during submission). But I'll have a think in case there's an elegant approach (maybe a data migration to add this one up front).

For ordering – I'll add a ticket to the backlog for custom ordering. For now I'll arrange the list in alphabetical order (of the shortname).

jits:
@Joseph-Irving – how do these sound to you?

- https://github.com/UKHomeOffice/platform-hub/pull/27/commits/9aa375050f10cd0b207b95595120350bff0e8090
- https://github.com/UKHomeOffice/platform-hub/pull/27/commits/e43f424283d20d828e50dd992ae64eb4988cc715

… any other text / help we need to add for this feature?

platform-hub-web/src/app/help/support/requests/support-requests-overview.html
```diff
@@ -13,6 +13,16 @@
 
     <h4 class="md-title">Submit a new support request</h4>
 
+    <p class="md-body-1" md-colors="{background: 'blue-grey-50'}">
+      <strong>Tip:</strong>
+      if you have a
+      <a ui-sref="identities">GitHub identity</a>
+      connected to the hub, we'll @ you in the support request issue
+      and you'll receive email notifications
+    </p>
```
Joseph-Irving: just notifications, that assumes they have email notifications set up

jits: Good point. Will change.

jits:
@marcinc – ready for review, whenever you have a sec :)

platform-hub-api/app/controllers/concerns/agents_initializer.rb
marcinc: Sure. In this case maybe better to call them `git_hub_service_agent`? Feels clearer to me. Also this will map dir structure nicely.

PR #33, opened 2017-03-24
Kube tokens

marcinc:
TODOS:
- [x] Add controller spec
- [x] Present tokens in User identities 
- [x] Add tokens file generator
- [x] Add S3 sync (manually triggered)
- [x] Add bots token support
- [x] Add tokens encryption

Extra bits:
- [x] Token reclaim functionality (for end user)
- [x] Revoke token (admin)
- [x] Admin can review changes before syncing tokens to S3 for individual clusters.


jits:
Needs rebasing, when you get a chance :)

platform-hub-web/src/app/app.routes.js
```diff
@@ -36,6 +40,40 @@ export const appRoutes = function ($stateProvider, $urlRouterProvider, $location
...
+      abstract: true,
+      url: '/kubernetes-tokens',
+      template: '<ui-view></ui-view>'
+    })
+      .state('kubernetes-tokens.list', {
+        url: '/list/:userId',
+        component: KubernetesTokensList,
+        data: {
+          authenticate: true
+        },
```
jits: Need to add `rolePermitted: 'admin'` to this route.

platform-hub-web/src/app/kubernetes-tokens/kubernetes-tokens-form.component.js
```diff
@@ -0,0 +1,103 @@
...
+  template: require('./kubernetes-tokens-form.html'),
+  controller: KubernetesTokensFormController
+};
+
+function KubernetesTokensFormController($state, hubApiService, logger, $stateParams) {
+  'ngInject';
+
+  const ctrl = this;
+
+  const userId = $stateParams.userId || (ctrl.transition && ctrl.transition.params().userId);
```
jits: `$stateParams` is deprecated – probably best not to use it.

platform-hub-web/src/app/kubernetes-tokens/kubernetes-tokens-form.component.js
```diff
@@ -0,0 +1,103 @@
...
+    ctrl.isNew = !cluster;
+
+    loadUser();
+
+    ctrl.clusters = [
+      {id: 'production', name: 'Production cluster'},
+      {id: 'development', name: 'Development cluster'},
+      {id: 'ci', name: 'CI cluster'},
+      {id: 'ops', name: 'OPS cluster'}
+    ];
```
jits: (as mentioned before) These should probably be driven by config to the backend, and then exposed in the API. Although, probably too much / complex data to be passed as config, so we could use the new `HashRecord` table here, expose it in an API, and either manually populate these in the db or provide an admin UI to edit these.

I can help with this if you think that's a good idea.

platform-hub-web/src/app/kubernetes-tokens/kubernetes-tokens-form.component.js
```diff
@@ -0,0 +1,103 @@
...
+  function loadToken() {
+    ctrl.loading = true;
+    ctrl.tokenData = null;
+
+    hubApiService
+      .getKubernetesTokens(userId)
+      .then(tokens => {
+        ctrl.tokenData = tokens.find(t => {
+          return t.cluster === cluster;
+        });
```
jits: Can use [`_.find(…)`](https://lodash.com/docs/4.17.4#find) here to simplify this a little.

platform-hub-web/src/app/kubernetes-tokens/kubernetes-tokens-form.html
```diff
@@ -0,0 +1,83 @@
...
+      <md-content layout-padding>
+        <form name="$ctrl.kubernetesTokenForm" role="form" ng-submit="$ctrl.createOrUpdate()">
+
+          <md-input-container class="md-block">
+            <label for="cluster">Cluster:</label>
+            <md-select ng-model="$ctrl.tokenData.cluster" 
+              name="cluster"
+              required
+              placeholder="Select the cluster"
+              ng-disabled="!$ctrl.isNew && $ctrl.tokenData.cluster"
```
jits: Ah, you can't edit the cluster once a token has been created? (if so, is this also enforced in the backend?)

platform-hub-web/src/app/kubernetes-tokens/kubernetes-tokens-form.html
```diff
@@ -0,0 +1,83 @@
...
+            <label for="groups">Groups:</label>
+            <input
+              name="groups"
+              ng-model="$ctrl.tokenData.groups"
+              required
+              placeholder="Comma separated list of groups.">
+            <div ng-messages="$ctrl.kubernetesTokenForm.groups.$error">
+              <div ng-message="required">Groups are required.</div>
+            </div>
+          </md-input-container>
```
jits: (maybe not worth it) We could add some HTML5 validation here (using a regex) to check the text entered matches a comma-separated list. Or maybe it's fine to just enforce this on the backend?

platform-hub-web/src/app/kubernetes-tokens/kubernetes-tokens-list.component.js
```diff
@@ -0,0 +1,100 @@
+export const KubernetesTokensListComponent = {
+  template: require('./kubernetes-tokens-list.html'),
+  controller: KubernetesTokensListController
+};
+
+function KubernetesTokensListController(roleCheckerService, hubApiService, logger, $stateParams, $mdDialog) {
+  'ngInject';
+
+  const ctrl = this;
+  const userId = $stateParams.userId;
```
jits: Likewise, shouldn't be using `$stateParams` here either. (See https://ui-router.github.io/guide/ng1/migrate-to-1_0#stateparams-deprecation for details)

platform-hub-web/src/app/kubernetes-tokens/kubernetes-tokens-list.component.js
```diff
@@ -0,0 +1,100 @@
...
+
+  ctrl.searchSelectedUser = null;
+  ctrl.loading = true;
+  ctrl.isAdmin = false;
```
jits: I think this entire component is guaranteed to only be routed to if you're an admin (assuming you agree with my comment above in the routes). In which case it doesn't need to be aware of admin status at all and can assume it (should hopefully simplify things).

platform-hub-web/src/app/kubernetes-tokens/kubernetes-tokens-list.component.js
```diff
@@ -0,0 +1,100 @@
...
+  function init() {
+    loadAdminStatus();
+    if (userId) {
+      loadUser(userId);
+      loadKubernetesTokens(userId);
+    }
+  }
+
+  function filterKubernetesTokensByUser() {
+    if (!ctrl.isAdmin && !ctrl.isProjectManager) {
```
jits: Ohhh, we're allowing project managers access to this?

platform-hub-web/src/app/kubernetes-tokens/kubernetes-tokens-list.component.js
```diff
@@ -0,0 +1,100 @@
...
+      return;
+    }
+    loadUser(ctrl.searchSelectedUser.id);
+    loadKubernetesTokens(ctrl.searchSelectedUser.id);
```
jits: Both of these methods set `ctrl.loading` back to `false` when whichever one finishes first. Probably not a big deal (just means data flashes in). But a better approach is to wrap both API calls in a `$q.all` and reset the loading parameter on a `finally` handler on the overall promise. There should be an example of this elsewhere in the codebase, or I can help with this if needed.

platform-hub-web/src/app/kubernetes-tokens/kubernetes-tokens-list.html
```diff
@@ -0,0 +1,100 @@
...
+          </p>
+        </md-card-title-text>
+      </md-card-title>
+
+      <md-card-actions layout="row" layout-align="start center">
+        <md-button
+          class="md-primary"
+          aria-label="View and edit details for this token"
+          ui-sref="kubernetes-tokens.edit({userId: t.user.id, cluster: t.cluster})">
+          Edit
```
jits: Maybe should say "View and edit"?

platform-hub-web/src/app/kubernetes-tokens/kubernetes-tokens-list.html
```diff
@@ -0,0 +1,100 @@
...
+      </md-card-actions>
+    </md-card>
+
+  </md-content>
+
+  <md-content layout-padding ng-if="!$ctrl.loading && $ctrl.tokens.length == 0">
+    <md-card>
+      <md-card-title>
+        <md-card-title-text>
+          <span>
```
jits: If you add the class `none-text` here (something specific to our codebase) then it will render the text in a way which is consistent with other places where no results/items are found.

platform-hub-api/app/models/kubernetes_token.rb
```diff
@@ -0,0 +1,27 @@
+class KubernetesToken
+  include ActiveModel::Model
+  include ActiveModel::Serialization
+  include ActiveModel::Associations
+
+  CLUSTERS = [
+    'production',
+    'development',
+    'ops',
+    'ci'
```
marcinc: Agreed. Will move that out somewhere external.

platform-hub-api/app/controllers/kubernetes/tokens_controller.rb
```diff
@@ -0,0 +1,112 @@
+class Kubernetes::TokensController < ApiJsonController
+
+  before_action :kube_identity
+
+  # GET /:user_id
+  def index
+    authorize! :read, @identity
+    render json: Kubernetes::TokenService.from_identity(@identity)
```
marcinc: The purpose of this method was to extract and return array of tokens from kubernetes identity. Will change the name to be more specific.

platform-hub-api/app/models/kubernetes_token.rb
```diff
@@ -0,0 +1,27 @@
+class KubernetesToken
+  include ActiveModel::Model
+  include ActiveModel::Serialization
+  include ActiveModel::Associations
+
+  CLUSTERS = [
```
marcinc: I don't think it's currently possible to use enum in ActiveModel. Moving the list over to hash record as discussed. We can refactor down the line.

platform-hub-api/app/models/kubernetes_token.rb
```diff
@@ -0,0 +1,27 @@
...
+    'ops',
+    'ci'
+  ]
+
+  attr_accessor :user_id, :cluster, :token, :groups
+
+  validates_presence_of :user_id, :cluster, :token, :groups
+  validates_inclusion_of :cluster, in: CLUSTERS
+
+  belongs_to :user
```
marcinc: Good point.

platform-hub-web/src/app/shared/model/kubernetes-clusters.js
```diff
@@ -0,0 +1,33 @@
...
+        })
+        .finally(() => {
+          // Reuse the same promise for some time, to prevent smashing the API
+          $window.setTimeout(() => {
+            fetcherPromise = null;
+          }, apiBackoffTimeMs);
+        });
+    }
+    return fetcherPromise;
+  }
```
jits: Awesome consistency here :) :)

platform-hub-web/src/app/kubernetes-tokens/kubernetes-tokens-list.component.js
```diff
@@ -0,0 +1,92 @@
...
+  function loadUserAndTokens(user) {
+    ctrl.loading = true;
+
+    hubApiService
+      .getUser(user)
+      .then(user => {
+        ctrl.user = user;
+      })
+      .then(() => {
+        fetchKubernetesTokens();
```
jits: Likewise, can put this in the `then` block above.

platform-hub-web/src/app/kubernetes-tokens/kubernetes-tokens-list.component.js
```diff
@@ -0,0 +1,92 @@
...
+  ctrl.deleteToken = deleteToken;
+  ctrl.filterKubernetesTokensByUser = filterKubernetesTokensByUser;
+
+  init();
+
+  function init() {
+    KubernetesClusters
+      .refresh();
+
+    if (userId) {
```
jits: If no `userId` is present, will `ctrl.loading` ever be set to `false`?

platform-hub-web/src/app/shared/icons.factory.js
```diff
@@ -11,6 +11,7 @@ export const icons = function () {
     projects: 'group_work',
     supportRequests: 'feedback',
     text: 'short_text',
-    users: 'perm_identity'
+    users: 'perm_identity',
+    kubernetes: 'vpn_key'
```
jits: Could you add this in a lexically ordered place?

platform-hub-api/app/controllers/kubernetes/tokens_controller.rb
```diff
@@ -0,0 +1,75 @@
...
+    render json: Kubernetes::TokenService.tokens_from_identity(@identity)
+  end
+
+  # PATCH/PUT /:user_id/:cluster
+  def create_or_update
+    authorize! :manage, @identity
+    tokens, created_or_updated_token = Kubernetes::TokenService.create_or_update_token(
+      @identity,
+      params[:cluster],
+      params[:token][:groups]
```
marcinc: Requirement on groups presence exists in the UI so they should always be set, however, if groups aren't set validation will fail. I don't think we need to guard it further.

jits: Yeah, fair enough. Just thinking in terms of a standalone API, and guarding from bad API requests (either through bugs or maliciousness).

platform-hub-api/app/controllers/kubernetes/tokens_controller.rb
```diff
@@ -0,0 +1,75 @@
...
+      render json: created_or_updated_token
+    else 
+      render_model_errors @identity.errors
+    end
+  end
+
+  # DELETE /:user_id/:cluster
+  def destroy
+    authorize! :manage, @identity
+    tokens, _ = Kubernetes::TokenService.delete_token(@identity, params[:cluster])
```
marcinc: New kubernetes token requires identity as it's associated with it but can change that to take a token array instead and identity id when needed.

platform-hub-api/app/controllers/kubernetes/tokens_controller.rb
```diff
@@ -0,0 +1,75 @@
...
+  def find_identity
+    user = User.find(params[:user_id])
+    @identity = user.identity(:kubernetes)
+
+    if @identity.nil?
+      @identity = user.identities.create!(
+        provider: :kubernetes,
+        external_id: user.email,
+        data: {tokens: []}
+      )
```
marcinc: I don't think it's worth it. We only ever going to be bootstrapping kubernetes identity in one place.

platform-hub-api/app/models/kubernetes_token.rb
```diff
@@ -0,0 +1,22 @@
+class KubernetesToken
+  include ActiveModel::Model
+  include ActiveModel::Serialization
+  include ActiveModel::Associations
+
+  attr_accessor :identity_id, :cluster, :token, :groups
+
+  validates_presence_of :identity_id, :cluster, :token, :groups
+  validates_inclusion_of :cluster, in: -> {
+    HashRecord.find_by!(id: 'clusters', scope: 'kubernetes').data.map {|c| c['id']}
```
marcinc: Good point. Changed.

platform-hub-api/app/controllers/kubernetes/tokens_controller.rb
jits: That could work well.

platform-hub-api/app/controllers/kubernetes/tokens_controller.rb
jits: Yeah, fair enough.

platform-hub-api/app/serializers/identity_serializer.rb
```diff
@@ -8,4 +8,11 @@ class IdentitySerializer < ActiveModel::Serializer
     :created_at,
     :updated_at
   )
+
+  attribute :kubernetes_tokens, if: -> { object.provider.to_sym == :kubernetes }
+
+  def kubernetes_tokens
+    tokens = Kubernetes::TokenService.tokens_from_identity(object)
+    ActiveModel::Serializer::CollectionSerializer.new(tokens, each_serializer: KubernetesTokenSerializer)
+  end
```
marcinc: Moved kubernetes_tokens method into a block. For some reason serializer / each_serializer wouldn't work when passed as attribute arguments. Leaving rest as is.

jits: Are you sure you did `has_many …` ?

marcinc: Discussed offline.

platform-hub-api/config/routes.rb
```diff
@@ -55,6 +57,15 @@
 
     resource :app_settings, only: [ :show, :update ]
 
+    resources :kubernetes_clusters, only: [ :index ]
```
marcinc: 👍

platform-hub-api/app/controllers/kubernetes/clusters_controller.rb
```diff
@@ -0,0 +1,22 @@
+class Kubernetes::ClustersController < ApiJsonController
+
+  before_action :load_kubernetes_clusters_hash_record
+
+  skip_before_action :require_authentication, only: :index
```
jits: This makes the endpoint open publicly – is that intended?

platform-hub-api/app/controllers/kubernetes/clusters_controller.rb
```diff
@@ -0,0 +1,19 @@
...
+  # GET /kubernetes_clusters
+  def index
+    authorize! :read, :kubernetes_clusters
+    render json: @kubernetes_clusters.data
+  end
+
+  private
+
+  def load_kubernetes_clusters_hash_record
+    @kubernetes_clusters = HashRecord.find_or_create_by!(id: 'clusters', scope: 'kubernetes') do |r|
```
jits: Similar point here re: using `HashRecord.kubernetes.` (though no big deal!)

platform-hub-api/app/models/audit.rb
```diff
@@ -15,6 +15,7 @@ class Audit < ApplicationRecord
   scope :ascending, -> { reorder(version: :asc) }
   scope :descending, -> { reorder(version: :desc) }
   scope :by_action, -> (action) { where(action: action) }
+  scope :by_actions, -> (actions) { where("action IN (?)", actions) }
```
jits: I think you can just use the exact same `where` condition as the scope above (i.e. just `where(action: actions)` and AR will automagically convert it into an `IN` query :)

platform-hub-api/app/models/kubernetes_token.rb
```diff
@@ -0,0 +1,47 @@
...
+  end
+
+  def decrypted_token
+    ENCRYPTOR.decrypt(self.token)
+  end
+
+  protected
+
+  def encrypt_token
+    return unless decrypted_token.nil? # only encrypt if not already encrypted
```
jits: What if the value of the token is updated in memory (and thus needs to be encrypted again)? (I know you probably don't use the instance of this class like that anywhere, but it may be possible in the future, so this _could_ be a bug waiting to happen?)

platform-hub-api/app/services/kubernetes/static_token_service.rb
```diff
@@ -0,0 +1,82 @@
...
+    def create_or_update(cluster, kind, user_name, groups = [])
+      static_tokens = get_static_tokens(cluster, kind)
+
+      record = static_tokens.data.find do |t|
+        t['user'] == user_name.to_s
+      end
+
+      if record.nil?
+        token = Kubernetes::TokenService.generate_secure_random
+        record = Hashie::Mash.new(
```
jits: Does this need to be a `Hashie::Mash` instance? (doesn't look like it's used later on in the method, so the convenience provided by Mash is probably not worth it?)

platform-hub-api/app/services/kubernetes/static_token_service.rb
```diff
@@ -0,0 +1,82 @@
...
+          token: ENCRYPTOR.encrypt(token),
+          user: user_name,
+          uid: Kubernetes::TokenService.generate_secure_random,
+          groups: Kubernetes::TokenService.send(:cleanup, groups)
```
jits: May be better to just make that method public (and call it something like `cleanup_groups`)?

platform-hub-api/app/services/kubernetes/static_token_service.rb
```diff
@@ -0,0 +1,82 @@
...
+        )
+
+        static_tokens.data << record
+      else
+        record['groups'] = groups.present? ? 
+          Kubernetes::TokenService.send(:cleanup, groups) : []
```
jits: As above

platform-hub-api/app/services/kubernetes/static_token_service.rb
```diff
@@ -0,0 +1,82 @@
...
+      end
+
+      static_tokens.save!
+      "Created/updated #{kind} account for `#{user_name}` (token: #{ENCRYPTOR.decrypt(record['token'])})"
```
jits: One thing to consider with messages like this: if they are getting logged out to STDOUT then they'll show up in ELK, and thus we'll be leaking sensitive info. 

**The same issue applies for audits – for better or worse, [I've made it so that all Audit entry creations get logged out to STDOUT](https://github.com/UKHomeOffice/platform-hub/blob/master/platform-hub-api/app/models/audit.rb#L10) – we should probably rethink this?** (So far I've tried not to put anything too sensitive in those fields that do get logged out [e.g. `comment`], so maybe just need to check that we're not outputting anything sensitive in [the fields being logged to STDOUT](https://github.com/UKHomeOffice/platform-hub/blob/master/platform-hub-api/app/models/audit.rb#L21-L31))

platform-hub-api/app/services/kubernetes/static_token_service.rb
```diff
@@ -0,0 +1,82 @@
...
+    def delete_by_token(cluster, kind, token)
+      static_tokens = get_static_tokens(cluster, kind)
+
+      static_tokens.data.reject! do |t|
+        ENCRYPTOR.decrypt(t['token']) == token
+      end
+
+      static_tokens.save!
+      "Deleted #{kind} account with token `#{token.to_s}`"
+    end
```
jits: With operations like these, we should really consider pessimistic table level locking? As it's possible that two users claim tokens at the same time and cause data inconsistency.

(Note: I tried using Postgres' advisory locks in https://github.com/UKHomeOffice/platform-hub/pull/59 but they didn't seem to be working as expected in our set up, so avoid those for now, I think).

platform-hub-api/app/services/kubernetes/token_claim_service.rb
```diff
@@ -0,0 +1,75 @@
...
+    extend self
+
+    module Errors
+      class TokenNotFound < StandardError; end
+      class TokenAlreadyDefinedForCluster < StandardError; end
+      class TokenInvalid < StandardError; end
+      class TokenAlreadyClaimed < StandardError; end
+    end
+
+    def claim_token(user, token)
```
jits: Should wrap everything in here in a transaction to avoid inconsistent data on failure.

platform-hub-api/app/services/kubernetes/token_claim_service.rb
```diff
@@ -0,0 +1,75 @@
...
+        identity.id,
+        claim.cluster,
+        claim.data.groups,
+        claim.data.token
+      )
+
+      raise Errors::TokenInvalid, "User `#{user.email}` token invalid: #{created_token.errors.full_messages.join('; ')}" unless created_token.valid?
+
+      identity.data['tokens'] = tokens
+      identity.save
```
jits: Should this be `identity.save!`? (So that it throws an error if it's invalid, rather than silently fail).

platform-hub-api/app/services/kubernetes/token_claim_service.rb
```diff
@@ -0,0 +1,75 @@
...
+        user.identity(:kubernetes) || 
+        user.identities.create!(
+          provider: :kubernetes,
+          external_id: user.email,
+          data: {tokens: []}
+        )
+
+      if identity.data['tokens'].find {|t| t['cluster'] == claim.cluster}
+        raise Errors::TokenAlreadyDefinedForCluster, 
+              "User `#{user.email}` already have a token assigned for `#{claim.cluster}` cluster!"
```
jits: Maybe we should overwrite the token? There's a possibility (albeit small) where:
- user requests new token
- admin uses the hub to grant them a new token for that cluster
- user finds their old token, and tries to claim it

… maybe this isn't an important use case to cover? (The only negative impact I can think of is we end up with stale token in the static list, which we can always clean up later).

platform-hub-api/app/services/kubernetes/token_file_service.rb
```diff
@@ -0,0 +1,57 @@
...
+        STATIC_TOKEN_KINDS.each do |kind|
+          static_tokens(cluster.to_s, kind).each do |t|
+            i = HashWithIndifferentAccess.new(t)
+            row = [ENCRYPTOR.decrypt(i[:token]), i[:user], i[:uid]]
+            row << i[:groups].join(',') unless i[:groups].empty?
+            csv << row
+          end
+        end
+
+        Identity.kubernetes.all.each do |i|
```
jits: Worth batching this?

platform-hub-api/app/services/kubernetes/token_file_service.rb
```diff
@@ -0,0 +1,57 @@
...
+
+    private
+    
+    # Static `system` tokens contain the full list of all core system tokens in use.
+    # For example tokens used by kublet, controller-manager, etc. 
+    #
+    # Static `user` tokens contain the list of all user tokens currently in use.
+    #
+    # Static `robot` tokens contain all non-user specific tokens in use. 
+    # These can include various robots, CI tokens etc.
```
jits: This comment would be really useful before the `STATIC_TOKEN_KINDS` constant, above :)

platform-hub-api/app/services/kubernetes/token_file_service.rb
```diff
@@ -0,0 +1,57 @@
...
+    def static_tokens(cluster, kind)
+      unless STATIC_TOKEN_KINDS.include?(kind)
+        raise Errors::UnknownStaticTokensKind, "`#{kind}` kind not supported."
+      end
+      hr = HashRecord.kubernetes.find_by(id: "#{cluster}-static-#{kind}-tokens")
+      return hr.data if hr.present?
+      []
```
jits: Could simplify with `hr.try(:data) || []`.

platform-hub-api/app/controllers/kubernetes/revoke_controller.rb
```diff
@@ -0,0 +1,35 @@
+class Kubernetes::RevokeController < ApiJsonController
+
+  # POST /kubernetes/revoke
+  def revoke
+    authorize! :manage, :identity
```
marcinc: Revoke functionality is only ever going to be available to the admin. The difference between this and standard token management (via Identity) for the user is that an admin should be able to quickly lookup and revoke a given token without knowing which user that token is associated to. The only action end user can do for the time being is to "claim" their token.

platform-hub-api/app/services/kubernetes/token_lookup.rb
```diff
@@ -0,0 +1,56 @@
...
+            kind: kind,
+            user: nil
+          )
+        end
+      end
+      nil
+    end
+
+    def lookup_identities(token)
+      Identity.kubernetes.all.each do |i|
```
jits: Same point, re: batching here.

Also, is it worth making a more specific query here to retrieve the exact record required, rather than traversing? (just thinking in terms of scalability of this method if we have 000s of users (we may not ;-))

platform-hub-api/app/models/audit.rb
marcinc: Good spot!

platform-hub-api/app/services/kubernetes/static_token_service.rb
marcinc: Totally agree. I'll review of the stdout logging in this piece and make sure no user / sensitive data is ever there. We can always refer to Audit events if we need more details.

platform-hub-api/config/spring.rb
```diff
@@ -1,4 +1,7 @@
 %w(
+  .env
+  .env.local
+  .env.test
```
jits: Ah! Nice catch :)

platform-hub-api/lib/tasks/kubernetes.rake
```diff
@@ -0,0 +1,125 @@
...
+
+    desc "Delete kubernetes cluster from list of managed clusters"
+    task :delete, [:id] => [:environment] do |t, args|
+
+      raise "ERROR: Missing argument!" if args.id.blank?
+
+      puts Kubernetes::ClusterService.delete(args.id)
+    end
+
+    def clusters_config
```
jits: Is this being used within this file?

platform-hub-api/lib/tasks/kubernetes.rake
```diff
@@ -0,0 +1,125 @@
...
+        r.data = []
+      end
+    end
+
+  end
+
+  
+  namespace :static_token do
+
+    desc "Creates or updates static token - `groups` as semicolon separated string of group names: 'group1 group2'."
```
jits: Missing semicolon in the example string?

platform-hub-api/app/services/kubernetes/static_token_service.rb
marcinc: Good point. Will have a look on how to solve that issue.

platform-hub-api/lib/tasks/kubernetes.rake
```diff
@@ -0,0 +1,125 @@
...
+      end
+
+      key = "#{args.cluster}-static-#{args.kind}-tokens"
+
+      if data.present?
+        if HashRecord.kubernetes.exists?(id: key)
+          HashRecord.kubernetes.find_by!(id: key).destroy
+        end
+
+        HashRecord.kubernetes.create!(id: key, data: data)
```
jits: Above and this needs wrapping in transaction.

platform-hub-api/app/controllers/kubernetes/changeset_controller.rb
```diff
@@ -0,0 +1,29 @@
+class Kubernetes::ChangesetController < ApiJsonController
+
+  before_action :load_changeset
+
+  # GET /kubernetes/tokens-changeset/:cluster
```
marcinc: Initially I used more verbose controller / route names but decided to keep it shorter. Happy to change if you think it makes things much clearer?

jits: Nah, that's fine – it's consistent with the other controllers too, and all within the `kubernetes` namespace, so fine for now. (If/when we add Kubernetes functionality that clashes [e.g. changesets management of, say, pods, we can consider renaming then).

platform-hub-api/app/controllers/kubernetes/revoke_controller.rb
jits: Makes sense.

platform-hub-api/app/services/kubernetes/token_claim_service.rb
marcinc: I think that's actually quite possible scenario. Agree with your suggestion here. Let's allow user claiming their old token to override exiting one (if present) - in such case we should ensure that groups allocation allocated by the admin (in step 2 above) takes precedence over the groups in the old token.

jits: Good point!

jits:
Thinking about the S3 sync some more… it would be good to have some checks/balances in place here. For example, if a zero length CSV string is generated (for whatever reason) then we should probably not post it to S3 at all. I'm thinking in terms of unexpected failure scenarios which could wipe out the tokens and cause mayhem. What do you think? Any other scenarios where we might want to have extra checks before posting the CSV contents to S3?

platform-hub-web/src/app/kubernetes-tokens/kubernetes-tokens-sync.html
```diff
@@ -0,0 +1,91 @@
...
+          <md-card-title>
+            <md-card-title-text>
+              <p>
+                <strong>Changed made by:</strong> {{c.user_name}}
+              </p>
+              <p>
+                <strong>Description:</strong> {{c.comment}}
+              </p>
+              <p>
+                <strong>Date:</strong> {{c.created_at | date:'medium'}}
```
jits: May be worth using the longer date format we've started using recently (I think we discussed this on another PR recently) – [example](https://github.com/UKHomeOffice/platform-hub/blob/master/platform-hub-web/src/app/announcements/editor/announcements-editor-list.html#L72). The benefit is it normalised to UTC and is more verbose.

platform-hub-web/src/app/kubernetes-tokens/kubernetes-tokens-sync.html
```diff
@@ -0,0 +1,91 @@
...
+          </md-card-title>
+
+        </md-card>
+
+        <md-button
+          class="md-primary"
+          aria-label="Sync kubernetes tokens"
+          ng-click="$ctrl.syncTokens($ctrl.cluster, $event)">
+          Sync
+        </md-button>
```
jits: I guess this button (and the one below) should only be enabled if a valid cluster is specified? (i.e. `$ctrl.cluster` is set) If so, could add `ng-disabled="!$ctrl.cluster` to the buttons.

platform-hub-web/src/app/kubernetes-tokens/kubernetes-tokens.module.js
```diff
@@ -0,0 +1,17 @@
+import angular from 'angular';
+
+import {KubernetesTokensListComponent} from './kubernetes-tokens-list.component';
+import {KubernetesTokensFormComponent} from './kubernetes-tokens-form.component';
```
jits: Similar comment here and below about ordering alphabetically (sorry).

platform-hub-web/src/app/shared/hub-api/hub-api.service.js
```diff
@@ -450,4 +460,99 @@ export const hubApiService = function ($rootScope, $http, $q, logger, events, ap
...
+        return response.data;
+      })
+      .catch(response => {
+        logger.error(buildErrorMessageFromResponse(`Failed to create or update a "${data.cluster}" kubernetes token for ${user.id}`, response));
+        return $q.reject(response);
+      });
+  }
+
+  function getKubernetesTokensChangeset(cluster) {
+    return $http
```
jits: Need to check that `cluster` is not null or empty before making the API request.

platform-hub-web/src/app/shared/identities/identities-list.component.js
```diff
@@ -82,4 +84,32 @@ function IdentitiesListController($mdDialog, Identities, Me, _) {
...
+      .ariaLabel('Claim token')
+      .initialValue('')
+      .targetEvent(targetEvent)
+      .ok('Claim token')
+      .cancel('Cancel');
+
+    $mdDialog
+      .show(confirm)
+      .then(claimedToken => {
+        ctrl.busy = true;
```
jits: Need to add this to top of controller (definition/initialisation).

jits: I just realised it was already used before, but not defined – my bad.

platform-hub-web/src/app/shared/identities/identities-list.html
```diff
@@ -32,11 +36,51 @@
...
+        </div>
+
+        <md-button class="md-secondary md-raised" 
+          ng-click="$ctrl.claimKubeToken($event)"
+          aria-label="Claim existing kubernetes token">
+          Claim token
+        </md-button>
+
+        <md-button class="md-secondary md-raised" 
+          ng-click="$ctrl.showKubeTokens=($ctrl.showKubeTokens ? false : true)"
```
jits: Can simplify this to `$ctrl.showKubeTokens = !$ctrl.showKubeTokens`.

platform-hub-web/src/app/kubernetes-tokens/kubernetes-tokens-list.component.js
```diff
@@ -0,0 +1,127 @@
...
+        fetchKubernetesTokens();
+      });
+  }
+
+  function fetchKubernetesTokens() {
+    ctrl.tokens = [];
+
+    // const identity = _.find(ctrl.user.identities, i => {
+    //   return i.provider === 'kubernetes';
+    // });
```
jits: Can remove comment?

platform-hub-web/src/app/shared/model/kubernetes-clusters.js
```diff
@@ -0,0 +1,33 @@
+import angular from 'angular';
+
+export const KubernetesClusters = function ($window, hubApiService, apiBackoffTimeMs, logger, _) {
```
jits: `logger` isn't being used in this file, so can be removed.

platform-hub-api/app/services/encryptor_service.rb
```diff
@@ -0,0 +1,15 @@
...
+  end
+
+  def encrypt(plaintext)
+    @encryptor.encrypt_and_sign(plaintext)
+  end
+
+  def decrypt(encrypted_data)
+    @encryptor.decrypt_and_verify(encrypted_data)
+  rescue ActiveSupport::MessageVerifier::InvalidSignature
+    nil
```
marcinc: I don't think we need to. Unless we want to log every failed decryption when data didn't have valid signature?

jits: True, we probably don't, in the normal case. But in the abnormal case, e.g. a bug in the consumer of this code, we'd end up with potential null pointer exceptions (or handling `nil` values when there should be). But I guess at that point we can debug closer.

PR #40, opened 2017-04-19
New hub onboarding/setup flow

jits:
**TODO:**

- [X] Redirect users to this new page when they log in, if they haven't already completed it
- ~~Allow platform theme resources to be scoped based on the user's new scope~~ _will do a separate PR_
- ~~Allow support request templates to be scoped based on the user's new scope~~ _will do a separate PR_

### DIFF
--- a/platform-hub-api/app/models/user.rb
+++ b/platform-hub-api/app/models/user.rb
@@ -6,6 +6,8 @@ class User < ApplicationRecord
   audited descriptor_field: :email
   has_associated_audits
 
+  before_create :build_flags
+
   enum role: {
     admin: 'admin'
   }
@@ -20,6 +22,12 @@ class User < ApplicationRecord
     dependent: :delete_all
 
   has_many :projects, through: :memberships
+
+  has_one :flags,
+    class_name: 'UserFlags',
+    foreign_key: 'id',
+    autosave: true,
+    dependent: :destroy
 
   pg_search_scope :search,
     :against => :name,
@@ -40,6 +48,19 @@ class User < ApplicationRecord
 
   def revoke_admin!
     self.role = nil
+    self.save!
+  end
+
+  def ensure_flags
+    self.flags.presence || self.create_flags!(id: self.id)
+  end
+
+  def update_flag name, value
+    ensure_flags.send("#{name}=", value)
+  end
+
+  def update_flag! name, value
+    update_flag name, value
     self.save!
   end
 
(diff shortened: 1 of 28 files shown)


platform-hub-api/app/models/user.rb
```diff
@@ -6,6 +6,8 @@ class User < ApplicationRecord
   audited descriptor_field: :email
   has_associated_audits
 
+  before_create :build_flags
```
marcinc: I might be missing something but I don't think this callback is defined?

platform-hub-api/app/models/user.rb
```diff
@@ -21,6 +23,12 @@ class User < ApplicationRecord
 
   has_many :projects, through: :memberships
 
+  has_one :flags,
```
marcinc: `has_many`?

platform-hub-api/app/serializers/user_serializer.rb
```diff
@@ -1,12 +1,20 @@
...
-
-  attributes :enabled_identities
+  has_many :identities, if: :is_admin_or_own?
 
-  def enabled_identities
+  attribute :enabled_identities do
     object.identities.pluck(:provider)
   end
+
+  has_one :flags, if: :is_admin_or_own?, serializer: UserFlagsSerializer do
```
marcinc: `has_many`?

platform-hub-api/spec/factories/user_flags.rb
```diff
@@ -0,0 +1,5 @@
+FactoryGirl.define do
+  factory :user_flags do
+
```
marcinc: Body missing.

platform-hub-api/app/models/user.rb
jits: This uses a method provided by the `has_one` association (figured it's easier to reuse that directly than wrap it in another method).

platform-hub-api/app/models/user.rb
jits: A user has one flags object (i.e one record in the `UserFlags` table), as discussed a couple of days ago :)

platform-hub-api/app/serializers/user_serializer.rb
jits: (see reply on previous comment)

platform-hub-api/spec/factories/user_flags.rb
jits: Right now this factory doesn't need anything initialised; will delete it.

platform-hub-api/app/models/user.rb
marcinc: Ah, true :)

platform-hub-api/app/models/user.rb
marcinc: OK

platform-hub-api/spec/factories/user_flags.rb
marcinc: 👍

PR #44, opened 2017-04-24
API: handle expired auth tokens

jits:
I.e.: don't log as an error!


platform-hub-api/app/controllers/concerns/authentication.rb
```diff
@@ -24,7 +24,9 @@ def create_or_fetch_user token
     # - it is already verified by the auth proxy in front
     # - expiry has already been checked by the auth proxy in front (TODO: perhaps recheck here)
     begin
-      payload, _ = JWT.decode(token, nil, false)
+      payload, _ = JWT.decode token, nil, false
```
marcinc: I know brackets are optional but I still prefer them when there is more than one argument :)

jits: Even if the arguments are so simple? ;-)

PR #45, opened 2017-04-24
New services onboarding flow

jits:
Second step of the new user onboarding flow (first being the hub setup that was done recently).

Currently only supports onboarding to the GitHub org + main team. We can add other services in the future.

### DIFF
--- a/platform-hub-api/app/controllers/me_controller.rb
+++ b/platform-hub-api/app/controllers/me_controller.rb
@@ -1,6 +1,8 @@
 class MeController < ApiJsonController
 
-  # This controller only ever acts on the currently authenticated user,
+  include GitHubOnboardingHelpers
+
+  # This controller should only ever act on the currently authenticated user,
   # so we do not need to peform an authorization checks.
   skip_authorization_check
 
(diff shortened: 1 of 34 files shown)


platform-hub-api/app/controllers/me_controller.rb
```diff
@@ -40,6 +42,24 @@ def complete_hub_onboarding
     end
   end
 
+  def complete_services_onboarding
+    # Currently only supports GitHub onboarding
+    success = handle_onboard_github_request current_user, audit_context
+
+    current_user.update_flag :completed_services_onboarding, true
+    current_user.save  # Note: doesn't _really_ matter if this save fails – just means the flag won't get set
```
marcinc: Fair enough. I guess this flag then translates into something in the UI that user will be potentially presented / prompted with again?

jits: Yeah, exactly – the worse case is we have a bug and flag triggers the banner in the UI permanently (which we'll add in a separate PR), which would be annoying, but not critical.

PR #49, opened 2017-06-05
Admins can manage a global contacts list

jits:
To be used later for things like global announcements.

### DIFF
--- a/platform-hub-api/app/models/contact_list.rb
+++ b/platform-hub-api/app/models/contact_list.rb
@@ -1,0 +1,41 @@
+class ContactList
+
+  HASH_RECORD_KEY_POSTFIX = '_contact_list'.freeze
+  HASH_RECORD_KEY_POSTFIX_REGEX = /#{HASH_RECORD_KEY_POSTFIX}$/
+
+  class << self
+    def find id
+      key = "#{id}#{ContactList::HASH_RECORD_KEY_POSTFIX}"
+      hash_record = HashRecord.contact_lists.find_or_create_by!(id: key) do |r|
+        r.data = { email_addresses: [] }
+      end
+      ContactList.new hash_record
+    end
+  end
+
+  # Needed for active_model_serializers to work :(
+  alias :read_attribute_for_serialization :send
+
+  attr_reader :hash_record
+
+  def initialize hash_record
+    @hash_record = hash_record
+  end
+
+  def update params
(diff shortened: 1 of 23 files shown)


platform-hub-api/spec/controllers/contact_lists_controller_spec.rb
```diff
@@ -0,0 +1,97 @@
+require 'rails_helper'
+
+RSpec.describe ContactListsController, type: :controller do
+
+  describe 'GET #show' do
+    before do
+      contact_list_hash_record = create :contact_list_hash_record
+      @contact_list = ContactList.new contact_list_hash_record
```
marcinc: Could skip `before` block here and just use `let`.

platform-hub-api/app/models/contact_list.rb
```diff
@@ -0,0 +1,41 @@
...
+      key = "#{id}#{ContactList::HASH_RECORD_KEY_POSTFIX}"
+      hash_record = HashRecord.contact_lists.find_or_create_by!(id: key) do |r|
+        r.data = { email_addresses: [] }
+      end
+      ContactList.new hash_record
+    end
+  end
+
+  # Needed for active_model_serializers to work :(
+  alias :read_attribute_for_serialization :send
```
jits: Unfortunately so – when the serializer runs, it looks for a `read_attribute_for_serialization` method on the object (which for AR objects is automatically mixed in). It's yucky and unfortunate.

platform-hub-api/spec/controllers/contact_lists_controller_spec.rb
jits: May not be able to use `let` here since those are lazily evaluated (well, unless I use `let!`). Generally, I prefer `let`'s to be very simple data creation / test data, so this is a potential candidate, but since I needed intermediate state I figured we stick to the convention of instantiated the "primary" object in the `before` block and assigning to an instance variable (like we do for other controller spec cases).

platform-hub-api/app/models/contact_list.rb
marcinc: ah yep. makes sense now.

platform-hub-api/spec/controllers/contact_lists_controller_spec.rb
marcinc: sure.

PR #50, opened 2017-06-05
Initial API for showing and managing announcements

jits:
UI + delivery of announcements will be built on top of this.


jits:
@marcinc – thanks! Yep, I snuck two quick fixes to audits in here, but should've made a separate PR :)

PR #52, opened 2017-06-07
API: fix the update protections for the Announcement model

jits:


platform-hub-api/spec/models/announcement_spec.rb
```diff
@@ -32,6 +32,31 @@
...
+    end
+
+    it 'should still allow deletion' do
+      a = create :announcement, title: 'foo'
+      expect(a.title).to eq 'foo'
+
+      a.update status: :delivering
+
+      # Need to reload before we can destroy it
+      Announcement.find(a.id).destroy
```
marcinc: you could possibly just `a.reload`.

PR #55, opened 2017-06-09
Backend: integrated `delayed_job` for processing background jobs

jits:
https://github.com/collectiveidea/delayed_job/tree/v4.1.3

This will be used in upcoming work on:
- delivering announcements
- syncing Kube tokens

---

**TODO:**

- [x] Test this locally with a dummy job (just to make sure it works end to end)
  - ~~Maybe turn this into an integration test (if it doesn't take too long to do so)~~


marcinc:
Any particular reason why you chose DelayedJob?

jits:
@marcinc –

> Any particular reason why you chose DelayedJob?

Initially was going to use Sidekiq (as we had discussed), but I believe it can't use Postgres as it's backing store, and I wanted to avoid having to introduce Redis for this. Also, I think `delayed_job` works very well for our modest needs here.

marcinc:
Yeah, this is what I thought. I think not introducing extra dependencies (Redis in this case) is good. Let's stick to Delayed::Job. We can always swap backend for ActiveJob later if needed.

jits:
@marcinc – makes sense! (Good point re: ActiveJob – we should stick to using it's interfaces, so it's easy to switch later if needed).

PR #57, opened 2017-06-13
Delivery of announcements (email and Slack)

jits:
TODO:

- [x] Update README with new config options
- [x] Update README with new background job and announcements trigger (i.e. how to work with this in dev)
- ~~More tests?~~


jits:
No longer WIP – ready for review.

platform-hub-api/app/services/announcements_processor_service.rb
```diff
@@ -0,0 +1,84 @@
...
+      announcement.update! status: :delivering
+
+      unless announcement.deliver_to.blank?
+        process_for_email_delivery announcement
+        process_for_slack_delivery announcement
+      end
+
+      # At this point we have to assume that any delivery mechanism triggered has
+      # worked as expected
+      announcement.update! status: :delivered
```
marcinc: As discussed f2f I think this is OK for now but we might need to revisit if we need to ensure that announcements have been really delivered - in case we change / add more distribution mechanisms and are able to get delivery status.

platform-hub-api/app/mailers/announcement_mailer.rb
```diff
@@ -0,0 +1,14 @@
+class AnnouncementMailer < ApplicationMailer
+
+  include ActionView::Helpers::TextHelper
+
+  def announcement_email announcement, recipients
+    @announcement = announcement
```
marcinc: Maybe change that to `@announcement_text = announcement.text` for consistency and use it directly in template?

platform-hub-api/app/services/announcements_processor_service.rb
jits: Agreed – we'll probably need a `status` _per_ delivery, so we can potentially retry.

platform-hub-api/app/mailers/announcement_mailer.rb
jits: Good idea, will change this.

PR #58, opened 2017-06-16
Users can now see how many global announcements they have unread

jits:
… and can mark all as read.


platform-hub-web/src/app/announcements/global-announcements.html
```diff
@@ -4,6 +4,13 @@
       <h3>
         <span>Global Announcements</span>
       </h3>
+      <span flex></span>
+      <md-button
+        aria-label="Mark all global announcements as read"
+        ng-click="$ctrl.markAllRead()">
+        <md-icon>drafts</md-icon>
```
marcinc: Should this icon be defined in icons.factory?

jits: Good point! Will move.

PR #59, opened 2017-06-20
More robust handling of auth token <-> user accounts

jits:
- Instead of relying on the `sub` value from Keycloak we now rely on the `email` to match the auth token to the underlying user. Note: we still use the `sub` value (if provided) as our own internal ID.
- Better handling of race conditions for new users – the client makes multiple requests at the same time, all of which will try to create a new user account, so we handle errors and retry (where it will then load the already created user instead).
- Better flow in the UI after logging in – we ensure that a request to the `/me` API can complete fully before closing the window. This should ensure that a user account is properly created straight after an initial log in.


platform-hub-api/app/services/auth_user_service.rb
```diff
@@ -0,0 +1,35 @@
...
+            external_id: auth_token_payload['sub'],
+            external_username: auth_token_payload['preferred_username'],
+            external_name: auth_token_payload['name'],
+            external_email: auth_token_payload['email'],
+            data: auth_token_payload
+          )
+        end
+      end
+
+    rescue ActiveRecord::RecordNotUnique
```
marcinc: Does it mean that when user exists it will still retry up to 3 times?

jits: Not quite – the only scenario where this error will occur is when multiple requests all hit at the same time and all try to _create_ a new user in the db (because they all haven't found an existing one). As soon as the user exists in the db (for that auth identity) any subsequent calls to this code will find the existing user and not try to create a new one.

PR #60, opened 2017-06-21
Various tweaks (mostly UI and user facing)

jits:

### DIFF
--- a/platform-hub-web/src/app/announcements/editor/announcements-editor-list.html
+++ b/platform-hub-web/src/app/announcements/editor/announcements-editor-list.html
@@ -70,7 +70,7 @@
             Publish at:
           </span>
           <span>{{a.publish_at | date:'EEE MMM d y HH:mm UTC':'UTC'}}</span>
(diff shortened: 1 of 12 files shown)


platform-hub-web/src/app/announcements/editor/announcements-editor-list.html
```diff
@@ -102,12 +102,12 @@
           <span md-colors="{color: 'blue-grey-700'}">
             Created at:
           </span>
-          <span>{{a.created_at}}</span>
+          <span>{{a.created_at | date:'EEE MMM d y HH:mm UTC':'UTC'}}</span>
```
marcinc: I believe we use `| date:'medium'` elsewhere. Should we stick to it?

jits: We do use `medium` as the date format in the user's list, however that doesn't include the timezone and it doesn't normalise the time to UTC (nor does it include the day of the week), so I've started using this custom format in recent features. At some point we should probably consolidate and use the same format every (wrapping it in a custom filter to make it nicer to use).

PR #61, opened 2017-06-22
More customisable contact lists (for admins; used in announcements)

jits:
Admins can now fully manage these lists.


platform-hub-api/spec/controllers/contact_lists_controller_spec.rb
```diff
@@ -41,51 +91,106 @@
...
+      end
+
+      it_behaves_like 'an admin' do
+        context 'for a contact list that doesn\'t yet exist' do
+          it 'should create a new contact list with the ID specified' do
+            expect(ContactList.count).to eq 0
+            expect(Audit.count).to eq 0
+            put :update, params: put_params
+            expect(response).to be_success
+            expect(ContactList.count).to eq 1
```
marcinc: you could wrap the above with 
```
expect { put :update, params: put_params }.to change { ContactList.count }.by(1)
```

jits: I've avoided this style in the controller tests so far – rather I've used a structure like:

```
<assert preconditions>
<do thing>
<assert postconditions>
```

(not generally a fan of that block style since you can only express one change in that condition, unless you chain together lots of RSpec matchers, in which case it looks a bit messy :)

marcinc: Sure.

PR #63, opened 2017-06-26
[ACPBoard #374] Added unit tests for the loggerService util.

KashifSaadat:


platform-hub-web/src/app/shared/util/array-utils.service.spec.js
```diff
@@ -41,7 +41,7 @@ describe('arrayUtilsService', () => {
       describe('given empty sorted IDs', () => {
         const sortedIds = [];
 
-        it('should not return an empty list on the left and all items on the right', () => {
```
jits: I think the original text was right?

jits: Oops, ignore my comment! I'm in "opposite mode" today ;-)

PR #64, opened 2017-06-27
Terms of service flows

jits:
As an admin, I would like to define a Terms of Service for users of the hub.

As a user, I would like to agree to the Terms of Service before I start using the features in the hub (i.e. when I first log in), and I want a link to the terms of service page available to me.


jits:
> Should we support multiple TOS agreements?

I don't think there's a use case for that just now? In the future, it's likely that each project would have it's own ToS, but I'm not sure we would have multiple ToS agreements for the platform as a whole (people barely read one :D)

PR #65, opened 2017-06-28
Admins can now specify a "cost centre code" on projects

jits:
Currently only viewable by admins and project managers (for their own projects).


platform-hub-web/src/app/projects/projects-form.html
```diff
@@ -49,6 +51,14 @@
             </textarea>
           </md-input-container>
 
+          <md-input-container class="md-block">
+            <label for="cost_centre_code">Cost centre code:</label>
+            <input
```
marcinc: Should this be coming from some dict of possible values to choose from I think - as optional field i.e. allowing no value? Probably some validation on inclusion in the model should follow too?

marcinc: Sure. Thought we could lay the foundations for that now and have that ready but can do later as well.

PR #66, opened 2017-06-28
Feature flags

marcinc:


platform-hub-api/config/routes.rb
```diff
@@ -14,6 +14,9 @@
     post '/me/complete_services_onboarding', to: 'me#complete_services_onboarding'
     post '/me/global_announcements/mark_all_read', to: 'me#global_announcements_mark_all_read'
 
+    get '/feature_flags', to: 'feature_flags#index'
+    get '/feature_flags/:flag', to: 'feature_flags#show'
```
jits: Just to future proof, can do this as:

```
resources :feature_flags, only: [ :index, :show ]
```

That way we can easily add in the other actions for admins to manage flags (which I can help with in a later PR).

jits: Also, I can change this when we need it.

platform-hub-web/src/app/app.run.js
```diff
@@ -13,6 +13,11 @@ export const appRun = function ($rootScope, $transitions, $state, authService, l
       $rootScope.AppSettings = AppSettings;
     });
 
+  // User already looged in
```
jits: Maybe something like:

```
// Fetch feature flags now if user is logged in (otherwise they will be fetched later 
// when user does log in).
```

platform-hub-web/src/app/kubernetes-tokens/kubernetes-tokens-form.component.js
```diff
@@ -30,6 +30,10 @@ function KubernetesTokensFormController($state, hubApiService, logger, _, Kubern
     ctrl.isNew = !cluster;
     ctrl.loading = true;
 
+    if (!FeatureFlags.isEnabled(FeatureFlags.keys.kubernetesTokens)) {
+      $state.go('home'); // feature disabled
+    }
```
jits: I have an idea on how to do this at the routing level – let's pair on this when you're back.

jits: (The problem with doing this here is it's not guaranteed that the feature flags have been fetched from the server at this point).

platform-hub-web/src/app/layout/shell.html
```diff
@@ -124,8 +124,9 @@ <h1 flex>{{$ctrl.AppSettings.getAppTitle()}}</h1>
             <md-sidemenu-button
               ng-repeat="item in $ctrl.adminNavStates"
               ui-sref="{{item.state}}"
+              ng-if="!_.has(item, 'featureFlag') || (_.has(item, 'featureFlag') && $ctrl.FeatureFlags.isEnabled(item.featureFlag))"
```
jits: Can remove the `_.has(item, 'featureFlag')` bit as it'll be true by the time that part of the expression is evaluated.

platform-hub-web/src/app/shared/model/feature-flags.js
```diff
@@ -0,0 +1,43 @@
...
+
+  function refresh(force) {
+    if (force || _.isNull(fetcherPromise)) {
+      fetcherPromise = hubApiService
+        .getFeatureFlags()
+        .then(flags => {
+          angular.copy(flags, data);
+          return data;
+        });
+    }
```
jits: Need to do the API backoff thing?

platform-hub-api/app/services/feature_flag_service.rb
```diff
@@ -0,0 +1,43 @@
...
+    f = feature_flags
+    f.with_lock do
+      f.data.reject! {|k,_| k == flag.to_s}
+      f.save!
+      "Deleted feature flag"
+    end
+  end
+
+  def is_enabled?(flag)
+    feature_flags.data.with_indifferent_access[flag] || false
```
jits: Maybe worth doing `!!feature_flags.data.with_indifferent_access[flag]` to ensure you get a boolean back?

marcinc: Could do, but that should always be bool by default.

jits: Unless the key doesn't exist in the first place (which will be the case when first used (since we said the frontend would control the actual flag data).

jits: Actually, in which case the `|| false` bit will hit – so maybe the net effect is the same? (i.e. just want to make sure that no matter what, this method always returns a boolean).

marcinc: That's exactly what's going to happen. When flag doesn't exist it defaults to false, otherwise take the `bool` value of existing flag.

platform-hub-api/app/services/feature_flag_service.rb
```diff
@@ -0,0 +1,43 @@
+module FeatureFlagService
+  extend self
+
+  # Feature flags data is stored in HashRecord as:
+  # data: {
+  #   'your-feature-flag-name': Bool,
```
jits: Maybe this example should use underscores, to better represent how we'll actually be using this?

platform-hub-api/config/routes.rb
```diff
@@ -72,17 +74,19 @@
...
-      get '/tokens/:user_id', to: 'tokens#index'
-      put '/tokens/:user_id/:cluster', to: 'tokens#create_or_update'
-      patch '/tokens/:user_id/:cluster', to: 'tokens#create_or_update'
-      delete '/tokens/:user_id/:cluster', to: 'tokens#destroy'
-      get '/clusters', to: 'clusters#index'
-      get '/changeset/:cluster', to: 'changeset#index'
-      post '/sync', to: 'sync#sync'
-      post '/claim', to: 'claim#claim'
-      post '/revoke', to: 'revoke#revoke'
+    constraints lambda { |request| FeatureFlagService.is_enabled?(:kubernetes_tokens) } do
```
jits: We should probably add a similar check in the `Identity` serialiser, so that the API doesn't return any token info if the feature is disabled (i.e. [here](https://github.com/UKHomeOffice/platform-hub/blob/master/platform-hub-api/app/serializers/identity_serializer.rb#L12)). That way we can also remove the logic in the UI, since if the API is not returning the data, then nothing is shown. (Though am happy for that logic to stay in, to be on the defensive side).

platform-hub-api/spec/factories/hash_records.rb
```diff
@@ -43,5 +43,16 @@
       end
     end
 
+    factory :feature_flags_hash_record do
+      scope 'general'
+      id 'feature_flags'
+      data do
+        {
+          'some-flag': false,
+          'other-flag': true
```
jits: Similar point here (and in the spec above) – may be good to use underscores to better represent how we'll actually have the feature flags.

platform-hub-api/app/services/feature_flag_service.rb
marcinc: yeah, I'll adjust comment here.

platform-hub-api/spec/factories/hash_records.rb
marcinc: sure.

PR #69, opened 2017-06-29
[ACPBoard #385] API Dockerfile updates

KashifSaadat:
- Switched to HO centos-base image in quay.io
- Install psql dependencies for bundler packages
- Install ruby, bundler and continue build as normal

Aim is to hopefully close down / reduce most of the security vulnerabilities identified in quay.io scans.


platform-hub-api/Dockerfile
```diff
@@ -1,22 +1,40 @@
...
+    && curl -sSL https://rvm.io/mpapis.asc | gpg --import - \
+    && curl -L get.rvm.io | bash -s stable \
+    && yum clean all
+RUN /bin/bash -l -c "rvm requirements"
+RUN /bin/bash -l -c "rvm install 2.3.3"
+RUN /bin/bash -l -c "rvm use 2.3.3 --default"
+RUN /bin/bash -l -c "gem install bundler --no-ri --no-rdoc"
 
-COPY . .
+# Install postgresql for libmq dependency
```
jits: `libmq` – is that a typo?

platform-hub-api/Dockerfile
```diff
@@ -1,22 +1,40 @@
...
+RUN yum install -y postgresql-devel postgresql-libs postgresql-contrib
 
+# Create app user
+WORKDIR /app
 RUN groupadd -r app && useradd -r -g app app
 RUN chown -R app:app /app
```
jits: I guess you can remove this line now?

platform-hub-api/Dockerfile
KashifSaadat: Yep good spot!

platform-hub-api/Dockerfile
KashifSaadat: This one is still required, creates and sets the working directory for the app files to be copied to.

platform-hub-api/Dockerfile
```diff
@@ -1,22 +1,46 @@
...
 
+ENV RUBY_VERSION 2.3.3
 ENV RAILS_ENV production
 ENV RACK_ENV production
 
-COPY Gemfile Gemfile.lock ./
-RUN bundle install --jobs 20 --retry 5 --deployment --without development test --no-cache
+# Update kernel-headers package
+RUN yum install -y wget \
+    && wget ftp://ftp.pbone.net/mirror/ftp.centos.org/7.3.1611/virt/x86_64/xen-46/kernel-headers-4.9.31-27.el7.x86_64.rpm \
```
jits: Can we trust `pbone.net`? (It's been a while since I've used CentOS so I'm lacking knowledge here :))

platform-hub-api/Dockerfile
```diff
@@ -1,22 +1,46 @@
...
+    && rm kernel-headers-4.9.31-27.el7.x86_64.rpm \
+    && yum remove -y wget
 
-COPY . .
+# Update packages & Install RVM
+RUN yum update -y \
+    && yum install -y \
+       which gcc-c++ patch readline readline-devel zlib zlib-devel \
+       libyaml-devel libffi-devel openssl-devel make bzip2 autoconf \
+       automake libtool bison iconv-devel sqlite-devel nodejs \
```
jits: Do we definitely need `sqlite-devel` and `nodejs`?

platform-hub-api/Dockerfile
```diff
@@ -1,22 +1,46 @@
...
+    && curl -sSL https://rvm.io/mpapis.asc | gpg --import - \
+    && curl -L get.rvm.io | bash -s stable \
+    && yum clean all
+RUN /bin/bash -l -c "rvm requirements"
```
marcinc: Am not sure about RVM. Would personally prefer rbenv but I guess we won't be changing versions very frequently so could as well just install ruby version we require?

platform-hub-api/Dockerfile
```diff
@@ -1,22 +1,46 @@
...
+RUN /bin/bash -l -c "rvm install 2.3.3"
```
marcinc: Also use `RUN` with multiple commands to minimise number of layers if possible.

platform-hub-api/Dockerfile
KashifSaadat: I'll switch it out for a more legit CentOS source. :D

platform-hub-api/Dockerfile
KashifSaadat: Tried both rbenv and rvm. The setup for rbenv was more convoluted and complicated the build steps. I did have a working rbenv setup but ended up dropping it from history with a force push. >.<

marcinc: Would skipping rvm entirely and installing ruby 2.3.3 doable?

platform-hub-api/Dockerfile
KashifSaadat: I'll clean that up and push shortly.

platform-hub-api/Dockerfile
KashifSaadat: Apparently not :) Dropped both deps out and builds successfully.

platform-hub-api/Dockerfile
```diff
@@ -1,22 +1,42 @@
...
+    && curl -L get.rvm.io | bash -s stable \
+    && yum clean all
 
+RUN /bin/bash -l -c "rvm requirements \
+    && rvm install 2.3.3 \
+    && rvm use 2.3.3 --default \
+    && gem install bundler --no-ri --no-rdoc"
+
+# Install postgresql for libpq dependency
+RUN yum install -y postgresql-devel postgresql-libs postgresql-contrib
```
marcinc: Maybe worth combining `RUN` commands above?

KashifSaadat: Whoops missed that, cleaning up now!

jaykeshur:
Should probably base this on: https://github.com/UKHomeOffice/docker-ruby

KashifSaadat:
@jaykeshur I did want to use that image and gave it a go at first. The `ONBUILD` performs a bundle install, which fails on this project for one of the gems as it requires libpq (not available by default, we pull it in from the postgres packages being installed). Is there a way you can ignore previous `ONBUILD` commands?

jaykeshur:
@KashifSaadat no, I don't think ONBUILD can be ignored

PR #70, opened 2017-06-29
Adding a prototype favicon to be used for display

AwkwardBen:
This is only a temporary image to be added to the platform hub. A new image can be suggested instead and repo to add this favicon to can also be changed.


jits:
As discussed offline, the source code here needs to be as general as possible, and not contain anything specific to our use (as much as possible :)). We'll add the favicon in our deployment through some other means (e.g. using NGINX to proxy to a file).

Appreciate the effort though.

PR #71, opened 2017-07-03
[ACPBoard #378] Bugfix 500 Server Error deleting missing identities

KashifSaadat:
- Added additional route to show identity, returning a json object
- Added 'before_action' on delete and show identity, to first attempt a get and return 404 on failure
- Updated controller tests

### DIFF
--- a/platform-hub-api/app/controllers/me_controller.rb
+++ b/platform-hub-api/app/controllers/me_controller.rb
@@ -12,15 +12,20 @@ class MeController < ApiJsonController
 
   def delete_identity
     # We assume that the `service` param has been validated by the route constraints
-    current_user.identity(params[:service]).destroy
+    user_identity = current_user.identity(params[:service])
+    if user_identity.present?
+      user_identity.destroy!
 
-    AuditService.log(
-      context: audit_context,
-      action: 'delete_identity',
-      comment: "User '#{current_user.email}' removed their #{params[:service]} identity"
-    )
+      AuditService.log(
+        context: audit_context,
+        action: 'delete_identity',
+        comment: "User '#{current_user.email}' removed their #{params[:service]} identity"
+      )
 
-    head :no_content
+      head :no_content
+    else
+      raise ActiveRecord::RecordNotFound
+    end
   end
 
   def agree_terms_of_service
--- a/platform-hub-api/spec/controllers/me_controller_spec.rb
+++ b/platform-hub-api/spec/controllers/me_controller_spec.rb
@@ -42,6 +42,37 @@ RSpec.describe MeController, type: :controller do
     end
   end
 
+  describe '#delete_identity' do
+    it_behaves_like 'authenticated' do
+      before do
+        @identity = create(:identity,
+          user: current_user
+        )
+      end
+
+      context 'for an identity that exists' do
+        it 'should delete the identity successfully' do
+          expect(Audit.count).to eq 0
+          delete :delete_identity, params: {service: @identity.provider}
+          expect(response).to be_success
+          expect(current_user.identity(@identity.provider)).to be nil
+          expect(Audit.count).to eq 1
+          expect(Audit.first.action).to eq 'delete_identity'
+          expect(Audit.first.user.id).to eq current_user_id
+        end
+      end
+
+      context 'for a non-existent identity' do
+        it 'should fail and return a 404' do
+          delete :delete_identity, params: {service: 'kubernetes'}
+          expect(response).to have_http_status(404)
+          expect(json_response['error']['message']).to eq "Resource not found"
+          expect(Audit.count).to eq 0
+        end
+      end
+    end
+  end
+
   describe '#complete_hub_onboarding' do
     it_behaves_like 'unauthenticated not allowed'  do
       before do
--- a/platform-hub-api/db/structure.sql
+++ b/platform-hub-api/db/structure.sql
@@ -3,7 +3,7 @@
 --
 
 -- Dumped from database version 9.6.1
--- Dumped by pg_dump version 9.6.2
+-- Dumped by pg_dump version 9.6.3
 
 SET statement_timeout = 0;
 SET lock_timeout = 0;


platform-hub-api/spec/controllers/me_controller_spec.rb
```diff
@@ -42,6 +42,87 @@
     end
   end
 
+  describe '#show_identity' do
+    it_behaves_like 'authenticated' do
+      before do
+        current_user.identities.create!(
```
marcinc: use factory here instead:
```ruby
create(:identity, 
  external_id: '1234567',
  external_username: test_auth_payload.preferred_username,
  ...
)
```

platform-hub-api/spec/controllers/me_controller_spec.rb
```diff
@@ -42,6 +42,87 @@
...
+        expect(json_response['error']['message']).to eq "No such identity 'kubernetes' currently exists for this user"
+        expect(Audit.count).to eq 0
+      end
+    end
+  end
+
+  describe '#delete_identity' do
+    it_behaves_like 'authenticated' do
+      before do
+        current_user.identities.create!(
```
marcinc: Same here, best to use factory.

platform-hub-api/app/controllers/me_controller.rb
```diff
@@ -12,29 +10,23 @@ def show
...
-      logger.warn "User '#{current_user.email}' attempted to access an identity which does not exist: #{params[:service]}"
-      render_error "No such identity '#{params[:service]}' currently exists for this user", :not_found
-    end
-  end
-
   def delete_identity
     # We assume that the `service` param has been validated by the route constraints
-    @user_identity.destroy
+    user_identity = current_user.identity(params[:service])
+    if !user_identity.nil?
```
marcinc: you could also express it in slightly more idiomatic way
```ruby
if user_identity.present?
#or 
unless user_identity.nil?
```

platform-hub-api/app/controllers/me_controller.rb
```diff
@@ -12,15 +12,20 @@ def show
 
   def delete_identity
     # We assume that the `service` param has been validated by the route constraints
-    current_user.identity(params[:service]).destroy
+    user_identity = current_user.identity(params[:service])
+    if !user_identity.nil?
```
jits: `if user_identity.present?` is a slightly nicer way to write this.

platform-hub-api/app/controllers/me_controller.rb
marcinc: Alternatively you can call destroy on `user_identity` without checking for presence and rescue `ActiveRecord::RecordNotFound` exception.

platform-hub-api/spec/controllers/me_controller_spec.rb
```diff
@@ -42,6 +42,51 @@
...
+          external_id: '1234567',
+          external_username: test_auth_payload.preferred_username,
+          external_name: test_auth_payload.name,
+          external_email: test_auth_payload.email
+        )
+      end
+
+      it 'should delete an identity successfully if it exists for the current user' do
+        expect(Audit.count).to eq 0
+        delete :delete_identity, params: {service: 'github'}
```
jits: The provider `github` here just happens to match the provider of the factory object you created above; a potentially nicer way would be to do `{service: @identity.provider}` here and make sure you assign `@identity = <factory object creation>` above.

platform-hub-api/spec/controllers/me_controller_spec.rb
```diff
@@ -42,6 +42,51 @@
     end
   end
 
+  describe '#delete_identity' do
+    it_behaves_like 'authenticated' do
+      before do
+        create(:identity,
+          user: current_user,
+          external_id: '1234567',
```
jits: For the purpose of this test, you probably don't need to explicitly set this field and all the ones below it.

platform-hub-api/spec/controllers/me_controller_spec.rb
```diff
@@ -42,6 +42,51 @@
...
+        expect(Audit.count).to eq 0
+        delete :delete_identity, params: {service: 'github'}
+        expect(response).to be_success
+        expect(current_user.identity('github')).to be nil
+        expect(Audit.count).to eq 1
+        expect(Audit.first.action).to eq 'delete_identity'
+        expect(Audit.first.user.id).to eq current_user_id
+      end
+
+      it 'should gracefully fail to delete an identity that does not exist' do
```
jits: If you don't mind, could you restructure these test cases to be more in line with how we do 404 test cases elsewhere ([example](https://github.com/UKHomeOffice/platform-hub/blob/master/platform-hub-api/spec/controllers/projects_controller_spec.rb#L51-L73)).

platform-hub-api/spec/controllers/me_controller_spec.rb
```diff
@@ -42,6 +42,51 @@
...
+
+      it 'should gracefully fail to delete an identity if called twice' do
+        expect(Audit.count).to eq 0
+        delete :delete_identity, params: {service: 'github'}
+        expect(response).to be_success
+        expect(current_user.identity('github')).to be nil
+        expect(Audit.count).to eq 1
+        expect(Audit.first.action).to eq 'delete_identity'
+        expect(Audit.first.user.id).to eq current_user_id
+        delete :delete_identity, params: {service: 'github'}
```
jits: Likewise here, and the two other usages above – should use `@identity.provider` to couple the value to the actual factory object.

KashifSaadat:
Updated following comments, rebased off master and squished additional commits.

platform-hub-api/spec/controllers/me_controller_spec.rb
```diff
@@ -42,6 +42,37 @@
...
+          expect(current_user.identity(@identity.provider)).to be nil
+          expect(Audit.count).to eq 1
+          expect(Audit.first.action).to eq 'delete_identity'
+          expect(Audit.first.user.id).to eq current_user_id
+        end
+      end
+
+      context 'for a non-existent identity' do
+        it 'should fail and return a 404' do
+          delete :delete_identity, params: {service: 'kubernetes'}
```
jits: Just to make it super obvious, could do something like `service: 'unknown_service'`.

jits: (discussed offline – it's tricky to do what I suggested due this being a route level constraint, so okay as-is).

PR #74, opened 2017-07-06
[PlatformHub][Web] Contact List display bug

KashifSaadat:
Fixed a display bug where contact list emails are hidden if any record occurs more than once.


KashifSaadat:
Yup good plan, update pushed.

PR #75, opened 2017-07-06
Announcements tweaks

jits:

This is part of ongoing work to address user feedback.


platform-hub-api/spec/mailers/announcement_mailer_spec.rb
```diff
@@ -3,48 +3,89 @@
...
+    end
+
+    it 'queues the email' do
+      assert_emails 1 do
+        @email.deliver_now
       end
     end
+
+    it 'renders the email (both in text and html)' do
+      expect(@email[:from].value).to eq from
```
marcinc: can you use dot notation here for clarity (if possible) like you do below?

jits: Unfortunately not, due to https://github.com/rails/rails/issues/7602#issuecomment-263771279

PR #80, opened 2017-07-07
Api: kube clusters api only presents cluster id and description

marcinc:


platform-hub-api/app/controllers/kubernetes/clusters_controller.rb
```diff
@@ -5,14 +5,14 @@ class Kubernetes::ClustersController < ApiJsonController
   # GET /kubernetes_clusters
   def index
     authorize! :read, :kubernetes_clusters
-    render json: @kubernetes_clusters.data
+    render json: @kubernetes_clusters.data.map {|c| c.with_indifferent_access.slice(:id, :description)}
```
jits: Could make a special serializer for this?

marcinc: Since it's just array of hashes this should do?

jits: Ok seems fine for now

PR #86, opened 2017-07-10
Web: update FAQ

jits:


jits:
Can you think of any other entries we should have? Or any other "tips & tricks"?

PR #87, opened 2017-07-10
Handle duplicate token claim/revocation + UI tweaks

marcinc:


platform-hub-api/app/services/kubernetes/token_claim_service.rb
```diff
@@ -11,33 +11,39 @@ class TokenAlreadyClaimed < StandardError; end
     end
 
     def claim_token(user, token)
-      claim = lookup(token)
+      claims = lookup(token)
 
-      if claim.nil?
+      if claims.nil?
         raise Errors::TokenNotFound, "Token `#{token}` not found!"
```
jits: Should the same error be thrown if `claims` is empty?

platform-hub-api/app/services/kubernetes/token_lookup.rb
```diff
@@ -29,18 +31,20 @@ def lookup_static_tokens(token, kind = 'user')
           )
         end
       end
-      nil
+      found.any? ? found : nil
```
jits: May be nicer to just return an empty array and use that to determine whether token(s) have been found or not?

platform-hub-api/app/services/kubernetes/token_lookup.rb
```diff
@@ -51,8 +55,8 @@ def lookup_identities(token)
...
```
jits: Likewise here.

platform-hub-api/app/services/kubernetes/token_revoke_service.rb
```diff
@@ -8,35 +8,42 @@ class TokenNotFound < StandardError; end
     end
 
     def remove(token)
-      t = lookup_identities(token) || lookup_static_tokens(token, 'user') || lookup_static_tokens(token, 'robot')
+      tokens = lookup_identities(token) || lookup_static_tokens(token, 'user') || lookup_static_tokens(token, 'robot')
 
-      if t.nil?
+      if tokens.nil?
         raise Errors::TokenNotFound, "Token `#{token}` not found!"
```
jits: Similar question here re: empty array vs `nil`

platform-hub-api/app/services/kubernetes/token_lookup.rb
marcinc: I thought about that but this would make chaining lookups more fiddly as `[]` is truthy by definition.

jits: Could you not just use `result.any?` in the lookups?

(To elaborate on my rationale: I just generally find `nil`s to be very scary, and hard to trace, when things are going wrong (e.g. debugging a bug). Whereas an empty array is conceptually more accurate, and a bit more graceful (e.g. in some of the other code you can just `.each` the array without checking it first.)

marcinc: Agreed. Will change to conform to the same return type.

PR #90, opened 2017-07-11
User deactivate

marcinc:


platform-hub-api/app/controllers/users_controller.rb
```diff
@@ -49,6 +58,42 @@ def revoke_admin
...
+  rescue => e
+    render_error "User activation failed - #{e.class}: #{e.message}",
+                :unprocessable_entity and return
+  end
+
+  # POST /users/:id/deactivate
+  def deactivate
+    keycloak_agent_service.deactivate_user(@user)
+
+    @user.deactivate!
```
jits: Is it worth moving this logic into a separate service (e.g. `UserActivationService`) so that the two steps of Keycloak then local hub activation/deactivation are encapsulated in one place (and thus could be used through the console, or background jobs)?

platform-hub-api/app/controllers/users_controller.rb
```diff
@@ -49,6 +58,42 @@ def revoke_admin
...
+
+    AuditService.log(
+      context: audit_context,
+      action: 'make_inactive',
+      auditable: @user
+    )
+
+    head :no_content
+  rescue => e
+    render_error "User deactivation failed - #{e.class}: #{e.message}",
```
jits: Similar point here, as above.

platform-hub-api/app/serializers/user_serializer.rb
```diff
@@ -1,5 +1,5 @@
 class UserSerializer < BaseSerializer
-  attributes :id, :name, :email, :role, :last_seen_at
+  attributes :id, :name, :email, :role, :last_seen_at, :is_active
```
jits: Like `identities` below, we should probably make this an "admin or own" property?

jits:
We should probably only return activated users for the `GET /users/search/:q` endpoint? This is currently used for the admin Kubernetes tokens page, and for adding team members to a project. We definitely don't want to show deactivated users for the latter use case, but I'm not sure about the first use case (e.g. admins may still want to see the tokens for a deactivated user). What do you reckon?

platform-hub-api/app/controllers/users_controller.rb
jits: Also, when we deactivate a user, should we delete all their project memberships?

platform-hub-api/app/serializers/user_serializer.rb
jits: Having said that, maybe we still want to show the active/inactive state of users to other users, e.g. in the project team listing? (in which case we may need a bit of UI for that)

platform-hub-api/app/controllers/users_controller.rb
```diff
@@ -49,6 +58,42 @@ def revoke_admin
...
+
+    AuditService.log(
+      context: audit_context,
+      action: 'make_active',
+      auditable: @user
+    )
+
+    head :no_content
+  rescue => e
+    render_error "User activation failed - #{e.class}: #{e.message}",
```
jits: I notice that the Keycloak Agent service does return specific errors, in which case we should probably catch the individual errors, log them internally and respond with appropriate API response, like [we do elsewhere](https://github.com/UKHomeOffice/platform-hub/blob/3af3078f3e8a5bd6aa83de7820d4673bd960909b/platform-hub-api/app/controllers/identity_flows_controller.rb#L36-L45)?

platform-hub-api/app/services/agents/keycloak_agent_service.rb
```diff
@@ -0,0 +1,141 @@
...
+  class KeycloakAgentService
+
+    module Errors
+      class KeycloakIdentityMissing < StandardError; end
+      class KeycloakUserRepresentationMissing < StandardError; end
+      class KeycloakAccessTokenRequestFailed < StandardError; end
+      class KeycloakAccessTokenExpired < StandardError; end
+    end
+
+    attr_accessor :token, :token_expires_at
```
jits: Are these required to be exposed outside of this class?

platform-hub-api/app/services/agents/keycloak_agent_service.rb
```diff
@@ -0,0 +1,141 @@
...
+          {
+            client_id: @client_id,
+            client_secret: @client_secret,
+            username: @username,
+            password: @password,
+            grant_type: 'password'
+          }
+      rescue => e
+        Rails.logger.error "Couldn't obtain access token from Keycloak: #{e.class} - #{e.message}"
+        raise
```
jits: Should this raise `Errors::KeycloakAccessTokenRequestFailed`?

platform-hub-api/app/services/agents/keycloak_agent_service.rb
```diff
@@ -0,0 +1,141 @@
...
+        @token = JSON.parse(response.body)['access_token']
+        @token_expires_at = JSON.parse(Base64.decode64(@token.split('.').second))['exp']
+        @token
+      else
+        Rails.logger.error "Couldn't obtain access token from Keycloak. Response: #{response.body}"
+        raise Errors::KeycloakAccessTokenRequestFailed
+      end
+    end
+
+    def get_user_representation(user)
```
jits: Should this be a private method?

platform-hub-api/app/services/agents/keycloak_agent_service.rb
```diff
@@ -0,0 +1,141 @@
...
+      user_id = keycloak_identity.try(:user_id)
+
+      raise Errors::KeycloakIdentityMissing if user_id.nil?
+
+      begin
+        response = client.get "/auth/admin/realms/#{@realm}/users/#{user_id}",
+          {},
+          {Authorization: "Bearer #{bearer_token}"}
+      rescue => e
+        Rails.logger.error "Couldn't get user representation from Keycloak: #{e.class} - #{e.message}"
```
jits: It's very likely that our User IDs will be outdated (because of things like realm consolidation, etc.) Which is why [we recently changed the auth and user lookup to rely on the email address](https://github.com/UKHomeOffice/platform-hub/pull/59). I think we have two options here:

1. Make the ID more reliable and up to date, but updating the value whenever encounter a new one (this would be done during the auth check and user lookup), or
1. Make a query to Keycloak using the email address and not the ID.

What do you think?

platform-hub-api/app/services/agents/keycloak_agent_service.rb
```diff
@@ -0,0 +1,141 @@
...
+        raise
```
jits: Should this raise `KeycloakUserRepresentationMissing`?

platform-hub-api/app/services/agents/keycloak_agent_service.rb
```diff
@@ -0,0 +1,141 @@
...
+
+      if response.success?
+        JSON.parse(response.body)
+      else
+        Rails.logger.error "Couldn't get user representation from Keycloak. Response: #{response.body}"
+        raise Errors::KeycloakUserRepresentationMissing
+      end
+    end
+
+    def update_user(representation = {})
```
jits: Should this be a private method?

platform-hub-api/app/services/agents/keycloak_agent_service.rb
```diff
@@ -0,0 +1,141 @@
...
+      begin
+        response = client.put "/auth/admin/realms/#{@realm}/users/#{user_id}",
+          representation.to_json,
+          {
+            'Authorization': "Bearer #{bearer_token}",
+            'Content-Type': 'application/json'
+          }
+      rescue => e
+        Rails.logger.error "Couldn't update user representation in Keycloak: #{e.class} - #{e.message}"
+        raise
```
jits: Raise a specific error?

platform-hub-api/app/services/agents/keycloak_agent_service.rb
```diff
@@ -0,0 +1,141 @@
...
+      retries ||= 0
+      representation = get_user_representation(user)
+      representation['enabled'] = state
+      update_user(representation)
+    rescue Errors::KeycloakAccessTokenExpired
+      retry if (retries += 1) < 3
+    end
+
+    def token_not_expired?
+      remaining_time = token_expires_at - Time.now.to_i
```
jits: Is this a JWT token? If so, you can use the same JWT helpers as the auth concern.

platform-hub-api/spec/controllers/concerns/authentication_spec.rb
```diff
@@ -13,8 +13,8 @@ def index
 
   before do
     expect(@controller).to receive(:authenticated?).and_call_original
-    expect(@controller).to receive(:current_user).and_call_original
-    expect(@controller).to receive(:token_and_options).and_return([auth_token, nil])
+    allow(@controller).to receive(:current_user).and_call_original
+    allow(@controller).to receive(:token_and_options).and_return([auth_token, nil])
```
jits: Any particular reason to change these? Now they won't be explicitly checked as part of the spec.

platform-hub-api/spec/controllers/concerns/authentication_spec.rb
```diff
@@ -35,14 +35,28 @@ def index
       test_auth_token
     end
 
-    it 'should check for a current user and create one' do
-      user = instance_double('User')
+    context 'when user is active' do
+      let(:user) { build(:user) }
```
jits: Is `is_active` set to `true` by default in the factory? I know the table column is defaulted to `true`, but it may be worth explicitly setting this in the factory too? (though very minor!)

platform-hub-web/src/app/users/users-list.html
```diff
@@ -72,6 +72,20 @@
...
+        </md-button>
+        <md-button
+          ng-if="!u.is_active"
+          class="md-primary"
+          aria-label="Activate user"
+          ng-click="$ctrl.activateUser(u)">
+          Activate
+        </md-button>
       </md-card-actions>
     </md-card>
```
jits: We should prominently show whether the user is active or not in the listings AND on the user page (assuming we have one). Maybe have a look at how we do badges in the announcements listings and do something similar here?

platform-hub-api/app/controllers/users_controller.rb
marcinc: Sure, we can wrap that in separate service. We want to leave project memberships unchanged in case user gets reactivated.

platform-hub-api/app/services/agents/keycloak_agent_service.rb
marcinc: I think keycloak api expects `user_id` explicitly, so retrieving user representation by email is most likely not possible. In this case the only thing we can do is to keep user_id in sync.

jits: Fair enough; we should probably do that in this PR then – probably after the point we look up the user for auth – we already touch the `last_seen_at` field, so we can update the `id` field at that point (need to bear in mind row level locking + cascading the ID update to any foreign keys).

platform-hub-api/app/services/agents/keycloak_agent_service.rb
jits: Ping

platform-hub-api/app/services/agents/keycloak_agent_service.rb
jits: Ping

platform-hub-api/app/services/agents/keycloak_agent_service.rb
marcinc: Not necessarily. As we might want to use it for other bits later. Can make it private - not fussed.

platform-hub-api/app/services/agents/keycloak_agent_service.rb
marcinc: Can make it private for now.

platform-hub-web/src/app/users/users-list.html
marcinc: Labels added. Relevant user actions disabled for deactivated users.

platform-hub-api/app/serializers/user_serializer.rb
marcinc: Added some UI. Deactivated users clearly labeled and relevant actions disabled.

platform-hub-api/app/models/user.rb
```diff
@@ -69,4 +80,17 @@ def update_flag! name, value
...
   end
 
+  def touch_and_update_main_identity payload
+    self.with_lock do
+      self.touch :last_seen_at
+    end
+
+    i = main_identity
+    i.with_lock do
+      if i.present? && i.external_id != payload['sub']
```
jits: `i.present?` check probably needs to go before the `i.with_local` (otherwise there may be a null pointer error, I think).

platform-hub-api/db/structure.sql
```diff
@@ -243,7 +243,7 @@ CREATE TABLE platform_themes (
     slug character varying NOT NULL,
     description text NOT NULL,
     image_url character varying NOT NULL,
-    colour character varying,
+    colour character varying NOT NULL,
```
jits: I'm not sure where this change came from?

platform-hub-api/spec/support/user_activation_helpers.rb
```diff
@@ -0,0 +1,53 @@
+module UserActivationHelpers
+
+  RSpec.shared_examples "verify responses to exceptions" do |action|
```
jits: The `verify responses to exceptions` bit might be a bit too generic? Could say something like `verify responses to exceptions for user activation`?

platform-hub-web/src/app/app.run.js
```diff
@@ -28,7 +28,29 @@ export const appRun = function ($q, $rootScope, $transitions, $state, authServic
...
     }
   });
+
+  // Listen for auth unauthorized event
+  const authUnauthorizedHandler = $rootScope.$on(events.auth.unauthorized, () => {
+    logger.error('User unauthorized.');
+    $state.go('home');
+  });
+
+  // Lister for auth forbidden event
```
jits: Tyop

platform-hub-web/src/app/app.run.js
```diff
@@ -28,7 +28,29 @@ export const appRun = function ($q, $rootScope, $transitions, $state, authServic
...
+  const authForbiddenHandler = $rootScope.$on(events.auth.forbidden, () => {
+    logger.error('Access forbidden.');
+    $state.go('home');
+  });
+
+  // Lister for auth deactivated event
```
jits: Tyop

platform-hub-web/src/app/kubernetes-tokens/kubernetes-tokens-form.html
```diff
@@ -2,9 +2,10 @@
   <md-toolbar md-scroll-shrink>
     <div class="md-toolbar-tools">
       <h3>
-        <span ng-if="$ctrl.isNew">New Token</span>
+        <span ng-if="$ctrl.isNew && $ctrl.user.is_active">New Token</span>
```
jits: This just controls the display of this title; maybe we should assume that you can't even make it to this form for a deactivated user? (i.e. add a check when the form loads up, and redirect to home). Unless I'm missing something and an admin should need to see this form for deactivated users? (in which case maybe we need an alternate title, or maybe the badge below is enough).

platform-hub-web/src/app/projects/projects-detail.html
```diff
@@ -115,27 +116,31 @@ <h4 class="md-subhead centred">Team members</h4>
...
                     <md-button
+                      ng-if="m.user.is_active"
                       ng-click="$ctrl.userOnboardGitHub(m.user.id, $event)"
                       aria-label="Onboard this user to the GitHub org and team">
                       Onboard GitHub
                     </md-button>
                   </md-menu-item>
                   <md-menu-item ng-if="$ctrl.allowOnboardOrOffboardGitHub(m)">
                     <md-button
+                      ng-if="m.user.is_active"
```
jits: If all these actions in the menu are turned off when the user is not active, then maybe we should just turn off the the whole menu in that case? (unless there are specific actions the PM/admin needs to be able to take on deactivated users?)

platform-hub-api/db/structure.sql
marcinc: Not sure either. I guess migrating db generates those - maybe minor version changes behave in different way.

platform-hub-web/src/app/kubernetes-tokens/kubernetes-tokens-form.html
marcinc: I purposely allowed access to deactivated users tokens page so that admin can delete individual tokens. Title at the top will change accordingly by showing "deactivated" badge so this should be good enough?

platform-hub-web/src/app/projects/projects-detail.html
marcinc: PM/admin should be still able to remove user from the project.

platform-hub-web/src/app/kubernetes-tokens/kubernetes-tokens-form.html
jits: Ah yeah – that sounds good.

platform-hub-web/src/app/projects/projects-detail.html
jits: Makes sense.

jits:
LGTM :shipit: :)

PR #93, opened 2017-07-18
Added fix for api alpine image to communicate with SSL endpoints

KashifSaadat:


.drone.yml
```diff
@@ -77,6 +77,8 @@ pipeline:
       - apk update && apk upgrade
       - apk --update add ruby ruby-dev ruby-io-console ruby-json ruby-bigdecimal tzdata
       - apk --update add postgresql-client postgresql-dev build-base libxml2-dev libxslt-dev libffi-dev
+      - apk --update add ca-certificates openssl
```
jits: I'm not 100% sure if it matters, but maybe do this before installing Ruby? I believe Ruby will link to the OpenSSL binaries, so may be best to update those first before Ruby (it probably doesn't matter as the linking will be to the latest `.so` anyway, but just to be safe :D).

KashifSaadat:
Cheers, just re-ordered the package installations so openssl stuff is done first.

PR #94, opened 2017-07-18
Web: fix formatting for description fields

jits:
We now better handle paragraphs of text that contain newlines.


KashifSaadat:
Tested multiple areas where I could specify descriptions / larger body of text, all appear to show better formatting / display paragraphs appropriately.

LGTM.

jits:
@KashifSaadat – thanks for testing this out! :)

PR #95, opened 2017-07-18
DRY up logic and UI for defining form fields

jits:
For reuse in an upcoming feature we're currently working on.


platform-hub-api/app/models/concerns/validate_hashes.rb
```diff
@@ -0,0 +1,111 @@
...
+
+      validate :validate_hashes_validation
+
+      # Check that config provided is valid
+      config_ok = config.is_a?(Hash) &&
+        config.keys.any? &&
+        config.values.all? { |h| (h.keys - CONFIG_ENTRY_ALLOWED_KEYS).empty? }
+
+      unless config_ok
+        raise 'Invalid config provided to validate_hashes'
```
marcinc: Maybe we should throw specific exceptions with message?

platform-hub-api/app/services/form_fields_service.rb
```diff
@@ -0,0 +1,43 @@
...
+
+  def all_types
+    ALL_ALLOWED_TYPES
+  end
+
+  def validate_types types
+    if types.proper_subset?(ALL_ALLOWED_TYPES)
+      types
+    else
+      raise('Invalid form field type(s) specified')
```
marcinc: Probably worth throwing specific error here too?

platform-hub-api/app/models/concerns/validate_hashes.rb
jits: Given this only runs at initial load time (i.e. when models are first loaded in memory), I figured it didn't need a specific exception, especially as it's a halting error (i.e. stops the server from loading at all).

platform-hub-api/app/services/form_fields_service.rb
jits: Same thing as above.

platform-hub-api/app/models/concerns/validate_hashes.rb
marcinc: Makes sense.

jits:
Cheers @marcinc !

PR #96, opened 2017-07-21
Initial support for announcement templates

jits:
Please see individual commits. This covers: admins being able to manage announcement templates, plus formatting and previewing of these templates.

An upcoming PR will integrate these templates into the process of creating announcements.

### DIFF
--- a/platform-hub-api/app/controllers/announcement_templates_controller.rb
+++ b/platform-hub-api/app/controllers/announcement_templates_controller.rb
@@ -69,6 +69,13 @@ class AnnouncementTemplatesController < ApiJsonController
     render json: AnnouncementTemplate.form_field_types
   end
 
+  # POST /announcement_templates/preview
+  def preview
+    templates, data = params.require([:templates, :data])
+    results = AnnouncementFormatterService.format templates, data
+    render json: results
+  end
+
   private
 
   def find_announcement_template
--- a/platform-hub-api/app/controllers/announcement_templates_controller.rb
+++ b/platform-hub-api/app/controllers/announcement_templates_controller.rb
@@ -1,0 +1,88 @@
+class AnnouncementTemplatesController < ApiJsonController
+
+  before_action :find_announcement_template, only: [ :show, :update, :destroy ]
+
+  authorize_resource
+
+  # GET /announcement_templates
+  def index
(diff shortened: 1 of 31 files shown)


platform-hub-api/app/controllers/announcement_templates_controller.rb
```diff
@@ -0,0 +1,88 @@
...
+
+      render json: @announcement_template, status: :created
+    else
+      render_model_errors @announcement_template.errors
+    end
+  end
+
+  # PATCH/PUT /announcement_templates/1
+  def update
+    if @announcement_template.update(announcement_template_params)
```
marcinc: Should this be `update!` so exception is raised when template is invalid?

platform-hub-api/spec/controllers/announcement_templates_controller_spec.rb
```diff
@@ -0,0 +1,270 @@
...
+
+        let :announcement_template_ids do
+          @announcement_templates.sort_by(&:shortname).map(&:friendly_id)
+        end
+
+        it 'should return a list of all announcement templates' do
+          get :index
+          expect(response).to be_success
+          expect(json_response.length).to eq total_announcement_templates
+          expect(pluck_from_json_response('id')).to eq announcement_template_ids
```
marcinc: Do you always expect result in particular order? If not then maybe change matcher type.

platform-hub-api/app/controllers/announcement_templates_controller.rb
jits: We could use that approach, but so far for `update` "endpoints we've been using the check-boolean-else-return-the-errors-object" (so the `else` branch of this would return the errors with an appropriate status code). What do you think?

platform-hub-api/spec/controllers/announcement_templates_controller_spec.rb
jits: In this particular case we do have explicit ordering, so worth checking that explicitly (but as you suggest, in other index endpoints we don't care about ordering and use the "contains"  matcher instead).

platform-hub-api/app/controllers/announcement_templates_controller.rb
marcinc: ah, yes. let's stick to what we have.

PR #97, opened 2017-07-25
Admins can now use templates when making an announcement

jits:


marcinc:
Build is failing though.

jits:
> Build is failing though.

Ah! Good catch, will look into it.

PR #108, opened 2017-08-01
Better handling of announcements with no delivery targets

### LINKED ISSUE #101: [phub] Announcements created with no delivery targets remain in 'awaiting_delivery' state
**Steps to recreate:**

1. When logged into the platform hub with an admin account, select 'Announcements' -> 'New'
2. Fill in all fields, but leave the bottom section entirely blank (labelled `and announcements can be delivered outside of the hub`), so that it doesn't send email / slack notifications.
3. Hit 'Create' to fire off the announcement
4. View the announcement in the 'Announcements' tab

**Expected Behaviour:**
The 'Status' field should say something along the lines of 'delivered' / 'not_applicable' / 'viewable'. Maybe even drop off the 'Deliver to' field or put a similar message.

**Actual Behaviour:**
```
Deliver to:
Status: awaiting_delivery
```

Not sure if this is potentially adding additional load with any background jobs picking up things in `awaiting_delivery` state and trying to process them?

jits:
Fixes #101

### DIFF
--- a/platform-hub-api/app/services/announcements_processor_service.rb
+++ b/platform-hub-api/app/services/announcements_processor_service.rb
@@ -15,16 +15,20 @@ class AnnouncementsProcessorService
     @logger.info "Processing announcement #{announcement.id} - #{announcement.level} - '#{announcement.title}'"
 
     begin
+
       announcement.update! status: :delivering
 
-      unless announcement.deliver_to.blank?
+      if announcement.has_delivery_targets?
         process_for_email_delivery announcement
         process_for_slack_delivery announcement
+
+        # At this point we have to assume that any delivery mechanism triggered has
+        # worked as expected
+        announcement.update! status: :delivered
+      else
+        announcement.update! status: :delivery_not_required
       end
 
-      # At this point we have to assume that any delivery mechanism triggered has
-      # worked as expected
-      announcement.update! status: :delivered
     rescue => e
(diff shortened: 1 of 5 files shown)


platform-hub-api/app/services/announcements_processor_service.rb
```diff
@@ -15,16 +15,22 @@ def process announcement
     @logger.info "Processing announcement #{announcement.id} - #{announcement.level} - '#{announcement.title}'"
 
     begin
-      announcement.update! status: :delivering
 
-      unless announcement.deliver_to.blank?
-        process_for_email_delivery announcement
-        process_for_slack_delivery announcement
+      announcement.with_lock do
```
jits: Question for @marcinc and @Joseph-Irving – this now causes the whole deliver of the announcement to be in a transaction… so any failure will rollback the updates and make it so the announcement is picked up to be processed again. This means that announcements will now be retried regardless of how far they got in the previous (failed) attempt. Is this a good?

A part of me thinks this is useful for handling intermittent issues. Another part of me is concerned that if there's a bug, we'll keep retrying and partial sending of this announcement will occur each time (e.g. 1st -> it sends emails but fails on Slack, 2nd -> it sends email, again, but fails on Slack again… and so on). That's probably not a good thing?

jits: I'll rollback this particular change until we've had a chance to discuss it properly.

PR #113, opened 2017-08-04
Admins can now resend an existing published announcement

jits:
This change also includes (to address some side effects of allowing redelivery):
- better specs for the `AnnouncementsProcessorService`
- better logic for when to update the template definitions for an announcement (i.e. now it only updates them for non-published announcements)
- better specs for the `Announcement` model, for the abovementioned change

### DIFF
--- a/platform-hub-api/app/controllers/announcements_controller.rb
+++ b/platform-hub-api/app/controllers/announcements_controller.rb
@@ -1,6 +1,6 @@
 class AnnouncementsController < ApiJsonController
 
-  before_action :find_announcement, only: [ :show, :update, :destroy, :mark_sticky, :unmark_sticky ]
+  before_action :find_announcement, only: [ :show, :update, :destroy, :mark_sticky, :unmark_sticky, :resend ]
 
   skip_authorization_check :only => [ :global ]
   authorize_resource except: [ :global ]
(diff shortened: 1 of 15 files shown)


platform-hub-api/app/controllers/announcements_controller.rb
```diff
@@ -103,6 +103,19 @@ def unmark_sticky
     head :no_content
   end
 
+  # POST /announcements/:id/resend
+  def resend
+    if @announcement.mark_for_resend!
```
marcinc: Q: Is it possible to resend the same announcement multiple times?

platform-hub-api/spec/models/announcement_spec.rb
```diff
@@ -160,4 +182,140 @@
...
+        before do
+          @announcement = create :announcement_from_template, original_template_id: initial_template
+          @announcement.update! publish_at: now
+        end
+
+        context 'when the original template\'s definitions are updated' do
+          it 'should not sync the definitions on updates' do
+            initial_template.update! spec: updated_spec
+
+            expect(@announcement.template_data).not_to eq HashWithIndifferentAccess.new(updated_template_definitions)
```
marcinc: A matter of preference but could use a shortcut here and in other cases: 
```
updated_template_definitions.with_indifferent_access
```

platform-hub-api/app/controllers/announcements_controller.rb
jits: Yep, it should be. (Figured admins may want to do that for some big announcements)

platform-hub-api/spec/models/announcement_spec.rb
jits: Ah yeah that's much nicer!!

PR #120, opened 2017-09-08
Management of kube cluster config + robot tokens

jits:


platform-hub-api/app/controllers/kubernetes/clusters_controller.rb
```diff
@@ -16,4 +35,16 @@ def load_kubernetes_clusters_hash_record
     end
   end
 
+  def cluster_params
+    params.require(:cluster).permit(
+      :id,
+      :description,
+      :s3_region,
+      :s3_bucket_name,
+      :s3_access_key_id,
```
marcinc: Secrets will probably now leak in logs so we should do something to prevent that...

jits: I _think_ I have this covered because of the list I updated in https://github.com/UKHomeOffice/platform-hub/blob/master/platform-hub-api/config/initializers/filter_parameter_logging.rb – but will check again :)

marcinc: Cool. Once added LGTM.

PR #121, opened 2017-09-08
API: fix a dodgy test

jits:
I'm genuinely not sure how that test ever passed; it has been flakey lately in latest CI, which may have exposed the timing issues;

### DIFF
--- a/platform-hub-api/spec/services/announcements_processor_service_spec.rb
+++ b/platform-hub-api/spec/services/announcements_processor_service_spec.rb
@@ -8,7 +8,7 @@ describe AnnouncementsProcessorService, type: :service do
     let(:announcement_mailer) { class_double('announcement_mailer') }
     let(:slack_notifier) { instance_double('Slack::Notifier') }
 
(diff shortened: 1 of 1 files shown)


platform-hub-api/app/services/announcements_processor_service.rb
```diff
@@ -83,7 +83,7 @@ def email_addresses_for announcement
         []
       end
 
-    contact_lists = [announcement.deliver_to['contact_lists']].compact.flatten
+    contact_lists = Array(announcement.deliver_to['contact_lists'])
```
marcinc: this won't remove potential `nil` entries.

jits: I believe it does?

```ruby
[2.3.3p222]> Array(nil)
=> []
```

jits: Oh, I think I know what you mean:

```ruby
[2.3.3p222]> Array([nil])
=> [nil]
```

PR #122, opened 2017-09-11
UI improvements for Kube tokens management

jits:


marcinc:
Whilst you touch this things could you also make groups visible only when specified? 

Just a check: Does robot tokens flow create audit entries which then should be included in changelog before syncing?

jits:
> Whilst you touch this things could you also make groups visible only when specified?

Sure.

> Just a check: Does robot tokens flow create audit entries which then should be included in changelog before syncing?

Good catch; I am making audit entries (in the new robot tokens API), but I don't know if they are showing up in the log that gets shown before – will check that and fix if needed.

PR #123, opened 2017-09-15
Short lived token escalation

marcinc:


platform-hub-api/app/controllers/kubernetes/tokens_controller.rb
```diff
@@ -65,6 +65,43 @@ def destroy
...
+
+    @identity.with_lock do
+      @identity.data[:tokens] = tokens
+
+      if @identity.save!
+        AuditService.log(
+          context: audit_context,
+          action: 'escalate_kubernetes_token',
+          auditable: @identity,
+          data: { cluster: params[:cluster] },
```
jits: Is it worth putting the group + timing requested here?

platform-hub-api/app/jobs/privileged_token_expirer_job.rb
```diff
@@ -0,0 +1,50 @@
+class PrivilegedTokenExpirerJob < ApplicationJob
+  queue_as :tokens_expire
```
jits: Just to keep consistent, could you make this `tokens_expire` (and then update the `is_already_queued` method below too)?

platform-hub-api/app/jobs/privileged_token_expirer_job.rb
```diff
@@ -0,0 +1,50 @@
...
+
+  def perform
+    return unless FeatureFlagService.is_enabled?(:kubernetes_tokens)
+    
+    Identity.kubernetes.find_each(batch_size: IDENTITY_BATCH_SIZE) do |i|
+      should_update = false
+      deescalated_token_group_clusters = []
+
+      tokens = Kubernetes::TokenService.tokens_from_identity_data(i.data).each do |token|
+        next if token.expire_privileged_at.nil? || Time.parse(token.expire_privileged_at).future?
```
jits: I'm never sure of the differences between `Date`, `Time` and `DateTime` in Ruby, but I think `DateTime` may be the recommended class to use in Rails apps (e.g. see this discussion here: https://stackoverflow.com/questions/9138515/when-should-i-use-datetime-vs-date-time-fields-in-ruby-rails).

platform-hub-api/app/jobs/privileged_token_expirer_job.rb
```diff
@@ -0,0 +1,50 @@
...
+
+        # Remove ALL privileged groups from user token groups and reset expiration timestamp
+        token.groups = token.groups - Kubernetes::TokenGroupService.privileged_group_ids
```
jits: Could cache the value of `Kubernetes::TokenGroupService.privileged_group_ids` at the beginning?

platform-hub-api/app/jobs/privileged_token_expirer_job.rb
```diff
@@ -0,0 +1,50 @@
...
+        token.expire_privileged_at = nil
+
+        deescalated_token_group_clusters << token.cluster
```
jits: Only add the cluster if it's not in already? (otherwise dups)

platform-hub-api/app/jobs/privileged_token_expirer_job.rb
```diff
@@ -0,0 +1,50 @@
...
+          begin
+            i.data[:tokens] = tokens
+            i.save!
+
+            deescalated_token_group_clusters.each do |cluster_id|
+              AuditService.log(
+                action: 'deescalate_kubernetes_token',
+                auditable: i,
+                data: { cluster: cluster_id },
+                comment: "Privileged kubernetes tokens expired for `#{i.user.email}` in `#{cluster_id}` via background job."
```
jits: s/tokens/token/

platform-hub-api/app/services/kubernetes/token_service.rb
```diff
@@ -29,6 +29,24 @@ def create_or_update_token(data, identity_id, cluster, groups, token = nil)
...
+      existing_token = tokens.find {|t| t.cluster == cluster}
+
+      if existing_token
+        if existing_token.groups.nil?
+          existing_token.groups = [ privileged_group ]
+        else
+          existing_token.groups = (existing_token.groups << privileged_group).uniq
+        end
+        # Upper bound for privilege expiration set at 6h
+        existing_token.expire_privileged_at = [expire_in_secs, 21600].min.seconds.from_now
```
jits: Could store the value `21600` as a const at the top of file?

platform-hub-api/spec/controllers/kubernetes/groups_controller_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+          create(:kubernetes_groups_hash_record)
+        end
+
+        it 'returns the list of cluster-wide privileged kubernetes groups' do
+          get :privileged
+          expect(response).to be_success
+
+          group = json_response.first
+
+          expect(group['id']).to eq 'acp:all:user:cw-privileged'
```
jits: Right now this is assuming a value from the factory, which would break if we changed the factory. (In theory, default values provided by a factory are for fields you don't care about). Maybe set the groups explicitly above and check for an explicit value that's set? This way you can also add a non-privileged group too and check that isn't returned. (I.e. need to also check number of responses to make sure it's been filtered appropriately).

platform-hub-api/app/controllers/kubernetes/tokens_controller.rb
marcinc: not crucial as it's part of description but sure we could add that in.

jits: Yeah, just thinking in terms of writing a query to get all audits for escalation using a particular group.

platform-hub-api/app/jobs/privileged_token_expirer_job.rb
marcinc: Currently we support 1 token per cluster. Added guard to ensure uniqueness though.

jits: Ah yeah.

platform-hub-api/spec/controllers/kubernetes/groups_controller_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+        let(:privileged) { true }
+
+        before do
+          create(:kubernetes_groups_hash_record, data: [
+            {
+              id: group_id,
+              privileged: privileged,
+              description: 'some description'
+            }
+          ])
```
jits: Maybe also add a non-privileged group and check that it doesn't get returned?

PR #124, opened 2017-09-15
API: Upgrade to Ruby v2.3.5 for local development

jits:
~~_Note: I'll merge this AFTER https://github.com/UKHomeOffice/platform-hub/pull/123 so it doesn't disrupt that work_.~~

---

https://www.ruby-lang.org/en/news/2017/09/14/ruby-2-3-5-released/

~~Note: for the Docker image, we are currently dependent on the version of Ruby provided by Alpine v3.5 (which is currently at Ruby v2.3.3-r0 – https://pkgs.alpinelinux.org/packages?name=ruby&branch=v3.5&repo=main&arch=x86_64&maintainer=). We may need to upgrade to Ruby 2.4 to get the latest in production.~~

Note: for the Docker image, the base image we use (Alpine v3.5) has also upgraded to Ruby v2.3.5-r0 – https://pkgs.alpinelinux.org/packages?name=ruby&branch=v3.5&repo=main&arch=x86_64&maintainer=).


KashifSaadat:
What's the plan for this? Should we be looking at the v2.4 upgrade instead, as this will fail due to v2.3.5 not being available in the alpine image.

jits:
@KashifSaadat – good catch! Yeah, I'll close this PR and we'll look do a Ruby 2.4 upgrade instead.

jits:
@KashifSaadat – ha! Looks like they (Alpine) just updated the version of Ruby. So I'll merge this in now to make the build work again :)

PR #131, opened 2017-09-28
API: project authorization clean up + more specs

jits:


platform-hub-api/spec/controllers/projects_controller_spec.rb
```diff
@@ -402,6 +402,18 @@
         end
       end
 
+      context 'not an admin but is project manager but of a different project' do
```
marcinc: `manager but of` -> `manager of`

platform-hub-api/spec/controllers/projects_controller_spec.rb
```diff
@@ -464,13 +476,13 @@
         end
       end
 
-      context 'not an admin but is project team manager but of a different team' do
+      context 'not an admin but is project manager but of a different project' do
```
marcinc: `but of` -> `of`

platform-hub-api/spec/controllers/users_controller_spec.rb
```diff
@@ -370,13 +366,66 @@
...
+      end
+
+      it_behaves_like 'an admin' do
+
+        context 'when user does not have a GitHub identity connected' do
+          it 'should return a 400 Bad Request with an appropriate error message' do
+            expect(Agents::GitHubAgentService).to receive(:new).with(any_args).and_return(git_hub_agent_service)
+            expect(git_hub_agent_service).to receive(:offboard_user).with(@user).and_raise(Agents::GitHubAgentService::Errors::IdentityMissing)
+            post :offboard_github, params: { id: @user.id }
+            expect(response).to have_http_status(400)
```
marcinc: I'm wondering if 422 would be more appropriate here?

jits: The problem I had with using [422](https://httpstatuses.com/422) here is it implies that the input data was wrong in some way. Whereas, it's really that the internal state of the user is not compatible with this request, so I'd like to say to the client that they shouldn't have even made this request in the first place (without checking that the user already has a GitHub identity connected).

400 is a bit of a catchall and blunt instrument here, but I'm not sure there's a better code we can use here?

jits:
Nice catches there @marcinc – thanks! 👍 Updated :)

platform-hub-api/spec/controllers/users_controller_spec.rb
marcinc: yeah. leave that with 400 for now.

PR #133, opened 2017-10-03
Kube clusters refactor

marcinc:
After deploy task: 

`HashRecord.kubernetes.find_by!(id: "clusters").delete`


platform-hub-api/app/controllers/kubernetes/clusters_controller.rb
```diff
@@ -1,52 +1,42 @@
...
+    data[:name] = params[:name]  # Cluster name in URL takes precedence
 
     Kubernetes::ClusterService.create_or_update data
 
     AuditService.log(
       context: audit_context,
       action: 'update_kubernetes_cluster',
-      data: { id: params[:id] },
-      comment: "Kubernetes cluster '#{params[:id]}' created or updated by #{current_user.email}"
+      data: { name: data[:name] },
```
jits: What about old audit records? This change means it'll be harder to map old and new records after. Maybe that's okay though, given we're aiming to make a few breaking changes with this work – what do you think?

platform-hub-api/app/controllers/kubernetes/clusters_controller.rb
```diff
@@ -1,52 +1,42 @@
...
   # GET /kubernetes/clusters
   def index
     authorize! :read, :kubernetes_clusters
-    clusters = @kubernetes_clusters.data.map do |c|
-      c.with_indifferent_access.slice(:id, :description)
-    end.sort_by! do |c|
-      c[:id]
-    end
-    render json: clusters
+    clusters = Kubernetes::ClusterService.all_clusters.sort_by(&:name)
```
jits: Maybe just `all` instead of `all_clusters` (given it's already the `ClusterService`)?

platform-hub-api/app/controllers/kubernetes/clusters_controller.rb
```diff
@@ -1,52 +1,42 @@
...
 
     data = cluster_params.to_h
-    data[:id] = params[:id]  # ID in URL takes precedence
+    data[:name] = params[:name]  # Cluster name in URL takes precedence
 
     Kubernetes::ClusterService.create_or_update data
 
     AuditService.log(
       context: audit_context,
       action: 'update_kubernetes_cluster',
```
jits: We should consider changing this action now to be the regular CRUD actions that we do for all other AR based resources.

Also, need to set the `cluster` object in the Audit now as the `auditable` object.

platform-hub-api/app/controllers/kubernetes/clusters_controller.rb
jits: Having said that, we don't need this anymore since we can just set the name to be the descriptor for the `auditable` (and thus the Audit record) will contain the name in a separate column.

platform-hub-api/app/services/kubernetes/cluster_service.rb
```diff
@@ -2,83 +2,50 @@ module Kubernetes
   module ClusterService
```
jits: Do we need this service anymore? For "regular" AR models we just tend to call the basic CRUD operations directly in the controllers, as opposed to wrapping up the logic in services like this. (e.g. `ProjectsController`). May be more consistent to do it that way?

Note: it's probably a good idea to wrap any non-standard CRUD logic in services like this, but it doesn't look like we do anything out of the ordinary here?

jits: Also, this applies to the fetch/querying methods too – seems we should just be able to use the model directly in the controller?

platform-hub-api/app/controllers/kubernetes/clusters_controller.rb
```diff
@@ -1,52 +1,42 @@
 class Kubernetes::ClustersController < ApiJsonController
 
-  before_action :load_kubernetes_clusters_hash_record
-
   # GET /kubernetes/clusters
   def index
     authorize! :read, :kubernetes_clusters
```
jits: Should probably remove these explicit checks and use the `before_action` now.

platform-hub-api/app/services/kubernetes/cluster_service.rb
```diff
@@ -2,83 +2,50 @@ module Kubernetes
...
+    def all_clusters
+      KubernetesCluster.all
     end
 
-    def clusters_config_hash_record
-      HashRecord.kubernetes.find_or_create_by!(id: 'clusters') do |r|
-        r.data = []
-      end
+    def all_cluster_names
+      all_clusters.map(&:name)
```
jits: Can just do `KubernetesCluster.pluck(:name)` to be more efficient.

marcinc:
> Also, what's the plan for data migration?

Export what we have in HR and migrate data over to new model.

jits:
> Also, what's the plan for data migration?
>> Export what we have in HR and migrate data over to new model.

Ah – I saw [your first comment on the PR](https://github.com/UKHomeOffice/platform-hub/pull/133#issue-262396042) and got worried :)

I'm guessing that this won't affect the sync job?

Also, we should probably take the site down whilst we're doing this migration.

platform-hub-api/app/controllers/kubernetes/clusters_controller.rb
```diff
@@ -1,52 +1,69 @@
 class Kubernetes::ClustersController < ApiJsonController
 
-  before_action :load_kubernetes_clusters_hash_record
+  before_action :find_cluster, only: [ :show, :update ]
+
+  skip_authorization_check only: [ :index, :show ]
+  load_and_authorize_resource class: KubernetesCluster, :except => [:index, :show]
```
jits: We still want to authorise the `index` and `show` actions? Also, don't need to use `load_and_authorize_resource` here (unless that's the only one that supports a custom `class`?)

Could change these 4 lines to something like:

```ruby 
before_action :find_cluster, only: [ :show, :update ]

authorize_resource class: KubernetesCluster
```

(assuming that last line actually works).

platform-hub-api/app/controllers/kubernetes/clusters_controller.rb
```diff
@@ -1,52 +1,69 @@
...
+  def create
+    cluster = KubernetesCluster.new(cluster_params)
 
-    Kubernetes::ClusterService.create_or_update data
+    if cluster.save
+      AuditService.log(
+        context: audit_context,
+        action: 'create',
+        auditable: cluster,
+        comment: "Kubernetes cluster '#{cluster.name}' created or updated by #{current_user.email}"
```
jits: Generally, for simple CRUD actions like these, we don't add a comment, as it doesn't add anything extra that isn't already there in the `Audit` attributes (and also saves a little extra space in DB).

If you do leave the comment in, you should remove the "or updated".

platform-hub-api/app/models/kubernetes_cluster.rb
```diff
@@ -0,0 +1,43 @@
+class KubernetesCluster < ApplicationRecord
+
+  include FriendlyId
+  include Audited
+
+  before_save :downcase_name
+
+  audited descriptor_field: :name
+
+  friendly_id :name
```
jits: We don't restrict what goes in the `name` field, do we? We can either add a custom validation for this ([like we do for the `ContactList` model](https://github.com/UKHomeOffice/platform-hub/blob/d465ec5f59de446f91f8cc7a9a853de9de2363fb/platform-hub-api/app/models/contact_list.rb#L3)), OR we should do the following instead:

```ruby
# In migration, add:
t.string :slug, null: false
t.index :slug, unique: true

# Then, in model:
friendly_id :name, :use => :slugged
```

… in either case, we still need the below (since name can be edited?):

```ruby
def should_generate_new_friendly_id?
  name_changed? || super
end
```

platform-hub-web/src/app/kubernetes/kubernetes-clusters-form.component.js
```diff
@@ -57,7 +57,7 @@ function KubernetesClustersFormController($state, KubernetesClusters, logger) {
     ctrl.saving = true;
 
     KubernetesClusters
-      .createOrUpdate(ctrl.cluster.id, ctrl.cluster)
+      .createOrUpdate(ctrl.cluster.name, ctrl.cluster)
```
jits: I guess this now needs to be split into separate `createX` vs `updateX` code paths (and corresponding methods in `KubernetesClusters` and `hubApiService`? (Check out the way it's done in the projects form component).

platform-hub-api/app/controllers/kubernetes/clusters_controller.rb
marcinc: We should probably keep is similar to projects, so index and show don't have to be protected as they never expose sensitive data. Will check if the above works tho.

jits: Ah that's a good point – we will probably need these open for the projects self-serve work anyway.

platform-hub-api/db/migrate/20171001181648_create_kubernetes_clusters.rb
```diff
@@ -0,0 +1,17 @@
+class CreateKubernetesClusters < ActiveRecord::Migration[5.0]
+  def change
+    create_table :kubernetes_clusters, id: :uuid do |t|
+      t.string :name, null: false
+      t.string :description, null: false
```
jits: Just realised this should probably be `t.text :description, null: false`, to match how we configure other description fields.

platform-hub-api/README.md
```diff
@@ -112,39 +112,6 @@ To set up local email delivery, ensure you have the following env variables set
 
 **As usual, be careful when sending out emails from your local development environment!**
 
-### Kubernetes Tokens Management
```
jits: This header is still needed (for the sub section(s)) below.

jits:
Thinking about the migration some more… maybe this should be part of the existing migration step for creating the new table? I [did this for contact lists](https://github.com/UKHomeOffice/platform-hub/blob/d465ec5f59de446f91f8cc7a9a853de9de2363fb/platform-hub-api/db/migrate/20170621140022_create_contact_lists.rb#L21-L37) and it worked very well.

marcinc:
Added migration in migration. Have another look pls.

PR #137, opened 2017-10-12
Kube token refactor

marcinc:
- [x] TODO: Add migration to port data from HR to new model.

Post deploy: 
- [ ] Remove tokens data from kubernetes identities:
```ruby
Identity.kubernetes.each do |i|
  if i.data && i.data['tokens']
    i.data.delete('tokens')
    i.save
  end
end
```

- [ ] Remove static robot tokens for all defined clusters:
```ruby
KubernetesCluster.all.each do |c|
  HashRecord.kubernetes.delete("#{c.name}-static-robot-tokens")
end
```


platform-hub-api/app/serializers/kubernetes_token_serializer.rb
```diff
@@ -1,5 +1,31 @@
...
+      name: object.user.name,
+      email: object.user.email
+    }
+  end
+
+  attribute :description, if: -> { object.robot? } do
+    object.description
+  end
+
+  attribute :expire_privileged_at, if: -> { object.expire_privileged_at.present? } do
```
jits: Likewise here you probably don't need the block (I know this was in the previous code, but that may have been an oversight).

platform-hub-api/app/services/kubernetes/changeset_service.rb
```diff
@@ -2,29 +2,37 @@ module Kubernetes
...
+        audit_entities_by_cluster_and_auditable_type(cluster, auditable_type, actions).where("created_at > ?", since)
+      end.flatten
     end
 
     private
 
-    def audit_entities(actions, cluster)
+    def last_sync(cluster)
+      audit_entities_by_cluster(cluster, :sync_kubernetes_tokens).first.try(:created_at) || 1.year.ago.utc.to_s(:db)
+    end
```
jits: Given now that we automatically sync behind the scenes, the current changeset entries can be a bit misleading. I wonder if we should rethink this bit – we could just display all changeset entries for the last X days **and** also show the `sync_kubernetes_tokens` event (highlighted) in the UI, so it becomes more like a history page.

Don't have to do this now! We can do this in a separate PR, or maybe as part of a new Audit UI – just wanted to get your thoughts on it at this stage.

platform-hub-api/db/migrate/20171003130836_create_kubernetes_tokens.rb
```diff
@@ -0,0 +1,16 @@
...
+      t.string :kind, null: false
+      t.string :token, null: false
+      t.string :name, null: false
+      t.string :uid, null: false
+      t.text :groups, array: true, default: []
+
+      t.timestamps
+    end
+
+    add_index :kubernetes_tokens, :token, unique: true
```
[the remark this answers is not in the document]
jits: Also needs indexes for: `kind` and `kubernetes_cluster_id` (when we figure out where it went ;-))

platform-hub-api/db/migrate/20171012110416_import_kubernetes_tokens.rb
```diff
@@ -0,0 +1,94 @@
+class ImportKubernetesTokens < ActiveRecord::Migration[5.0]
```
jits: It's quite possible that the tokens sync job will run during this migration (if we're very unlucky). Wonder if we should somehow turn this off during the migration? E.g. we could turn off the `kubernetes_tokens` feature flag?

platform-hub-api/db/migrate/20171012110416_import_kubernetes_tokens.rb
```diff
@@ -0,0 +1,94 @@
...
+          puts "-- Token imported for user #{u.id} and cluster `#{cluster_name}`"
+        rescue => e
+          puts "-- Token import for user #{u.id} and cluster `#{cluster_name}` failed with #{e.message}"
+        end
+      end
+    end
+
+    puts "===== IMPORTING ROBOT TOKENS ====="
+
+    default_user = User.admin.first
```
jits: Should this be sorted by `created_at`?

platform-hub-api/app/controllers/kubernetes/tokens_controller.rb
```diff
@@ -1,127 +1,194 @@
...
+        @identity.user.robot_tokens.new(
+          data.merge(
+            name: token_params[:robot_name],
+            description: token_params[:robot_description]
+          )
+        )
+      when 'user'
+        @identity.tokens.new(
+          data.merge(
+            name: @identity.user.email
```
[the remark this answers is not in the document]
marcinc: Yes, for some users there will be no kubernetes identity initially so we need to ensure it's created. Currently individual kubernetes tokens are associated with identity.

platform-hub-api/app/controllers/kubernetes/tokens_controller.rb
```diff
@@ -1,127 +1,194 @@
 class Kubernetes::TokensController < ApiJsonController
```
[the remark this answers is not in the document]
marcinc: I don't think it's necessary really. Could extract some bits to private controller methods if that helps?

platform-hub-api/app/controllers/kubernetes/tokens_controller.rb
```diff
@@ -1,127 +1,194 @@
...
-          comment: "Kubernetes `#{params[:cluster]}` token removed for user '#{@identity.user.email}'"
-        )
+    @token.destroy
+
+    AuditService.log(
+      context: audit_context,
+      action: 'destroy',
+      auditable: @token,
+      data: {
+        cluster: @token.cluster.name
```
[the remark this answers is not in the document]
marcinc: It works just fine.

platform-hub-api/app/controllers/kubernetes/tokens_controller.rb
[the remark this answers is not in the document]
jits: Nah, we can always extract out later if it grows / is needed.

platform-hub-api/app/controllers/kubernetes/tokens_controller.rb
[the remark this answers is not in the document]
jits: Ah cool, makes sense.

platform-hub-api/app/controllers/kubernetes/tokens_controller.rb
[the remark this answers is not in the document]
jits: 👍

platform-hub-api/app/models/kubernetes_token.rb
```diff
@@ -1,18 +1,134 @@
...
+  def user
+    if robot?
+      tokenable
+    elsif user? # via identity
+      tokenable.user
+    end
+  end
+
+  def privileged?
+    !expire_privileged_at.nil?
```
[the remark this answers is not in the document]
marcinc: Good call. Will change.

platform-hub-api/spec/controllers/kubernetes/tokens_controller_spec.rb
```diff
@@ -42,201 +27,540 @@
...
+            updated = KubernetesToken.user.first
+            expect(updated.kind).to eq 'user'
+            expect(updated.name).to eq @token.name
+            expect(updated.cluster).to eq @token.cluster
+            expect(updated.groups).to match_array put_data[:token][:groups].split(",")
+            expect(Audit.count).to eq 1
+            # audit = Audit.first
+            # expect(audit.action).to eq 'update'
+            # expect(audit.auditable.id).to eq @token.id
+            # expect(audit.user.id).to eq current_user_id
```
jits: Needed?

platform-hub-api/spec/controllers/kubernetes/tokens_controller_spec.rb
```diff
@@ -42,201 +27,540 @@
...
+            expect(updated.kind).to eq 'robot'
+            expect(updated.name).to eq robot_name
+            expect(updated.description).to eq robot_description
+            expect(updated.cluster).to eq @token.cluster
+            expect(updated.groups).to match_array put_data[:token][:groups].split(",")
+            expect(Audit.count).to eq 1
+            # audit = Audit.first
+            # expect(audit.action).to eq 'update'
+            # expect(audit.auditable.id).to eq @token.id
+            # expect(audit.user.id).to eq current_user_id
```
jits: Also, needed?

platform-hub-api/db/migrate/20171012110416_import_kubernetes_tokens.rb
marcinc: Assumed that any admin will be good enough, but could get the very first one if that makes the difference?

jits:
This is an epic refactor :)

UI bits generally look good. My suggestion would be to make the methods in `hubApiService` mimic the API design/footprint as close as possible (e.g one function per endpoint), and then move the `user` vs `robot` token specific usage of the API into a `KubernetesToken` model (or split into `KubernetesUserToken` and `KubernetesRobotToken`) – what do you think?

platform-hub-api/db/migrate/20171012110416_import_kubernetes_tokens.rb
jits: Probably doesn't make a huge difference – fine as-is.

platform-hub-api/db/migrate/20171003130836_create_kubernetes_tokens.rb
```diff
@@ -0,0 +1,16 @@
+class CreateKubernetesTokens < ActiveRecord::Migration[5.0]
+  def change
+    create_table :kubernetes_tokens, id: :uuid do |t|
+      t.references :tokenable, polymorphic: true, type: :uuid, index: true
+      t.string :kind, null: false
+      t.string :token, null: false
+      t.string :name, null: false
+      t.string :uid, null: false
+      t.text :groups, array: true, default: []
+
```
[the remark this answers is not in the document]
marcinc: Migration was in a state of flux. Fixed now.

platform-hub-api/app/models/kubernetes_token.rb
```diff
@@ -1,18 +1,134 @@
...
+  audited descriptor_field: :name
 
-    new(params)
+  before_save :downcase_name
+
+  belongs_to :tokenable, polymorphic: true
+  belongs_to :cluster, class_name: KubernetesCluster, foreign_key: :kubernetes_cluster_id
+
+  scope :privileged, -> { where.not(:expire_privileged_at => nil) }  
+  scope :by_user, ->(u) {
```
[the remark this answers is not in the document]
marcinc: Not necessarily. `by_user` means all by that given user.

platform-hub-api/app/models/kubernetes_token.rb
```diff
@@ -1,18 +1,134 @@
...
+    format: {
+      with: NAME_REGEX,
+      message: "can contain only letters, numbers, underscores, dashes, dots and @"
+    }
+  
+  validate :token_must_not_be_blank
+  validate :one_user_token_per_cluster
+  validate :robot_name_unique_for_given_cluster
+  validate :robot_description_present
+
```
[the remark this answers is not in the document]
marcinc: Good point. I made those fields as readonly.

platform-hub-api/app/models/kubernetes_token.rb
```diff
@@ -1,18 +1,134 @@
...
+    ENCRYPTOR.decrypt(token)
+  end
+
+  def user
+    if robot?
+      tokenable
+    elsif user? # via identity
+      tokenable.user
+    end
+  end
```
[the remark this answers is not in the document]
marcinc: Added extra validations for it.

platform-hub-api/app/serializers/kubernetes_token_serializer.rb
```diff
@@ -1,5 +1,31 @@
-class KubernetesTokenSerializer < KubernetesTokenBaseSerializer
-  attribute :expire_privileged_at, if: -> { object.expire_privileged_at.present? }  do
+class KubernetesTokenSerializer < BaseSerializer
+  belongs_to :cluster
```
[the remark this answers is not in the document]
marcinc: Thought cluster description could come handy when presenting tokens to the user. If you think that's too much (?) I can change it.

platform-hub-api/app/services/kubernetes/changeset_service.rb
marcinc: Let's discuss and make changes in a separate PR.

platform-hub-api/db/migrate/20171012110416_import_kubernetes_tokens.rb
marcinc: Sure, we could turn the flag off and switch it back at the end of migration. Though token sync won't change token data so it should have no effect plus it's gonna be rather quick. Alternatively we can take it out of migration and run as standalone.

platform-hub-api/app/models/kubernetes_token.rb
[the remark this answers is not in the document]
jits: That's not clear from just `by_user` – I first read that as "just the ones associated directly with the user object" (as is usually the convention with naming scopes). `KubernetesToken.all_for_user(user)` makes it a bit clearer that it's all the tokens for that user, regardless of how that's derived.

Another way of looking at it: what if I wanted a scope that just gave me the tokens that are associated with a user object directly – what would that be called?

platform-hub-api/app/models/kubernetes_token.rb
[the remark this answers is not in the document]
jits: 👍

platform-hub-api/app/serializers/kubernetes_token_serializer.rb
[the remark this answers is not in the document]
jits: Yeah, we don't have a a good strategy at the moment for embedded objects, and how much of that resource to render. At the moment, I'm just conscious of returning X number of token resources, with probably the exact same cluster details within each. Having said that, this isn't a big deal at the moment so happy to leave as-is.

platform-hub-api/app/services/kubernetes/changeset_service.rb
jits: Cool 👍

platform-hub-api/db/migrate/20171012110416_import_kubernetes_tokens.rb
jits: The issue is if the tokens sync job is runs just before the migration kicks in, or halfway through, and then it writes out an incomplete tokens.csv file to S3 – that could screw things up in the cluster, I think?

jits: I think it's okay to have this migration script, but before we deploy this latest batch of work, we should turn off the feature flag, and only enable it once everything is migrated and ready. Just in case.

platform-hub-api/app/models/concerns/read_only_associations.rb
```diff
@@ -0,0 +1,40 @@
+module ReadOnlyAssociations
+  extend ActiveSupport::Concern
+
+  class_methods do
+
+    # Example
+    # read_only_associations :foo, :bar, {baz: :polymorphic}
+    def read_only_associations *associations
```
jits: Is this actually used anywhere?

jits: Also, I don't think you need this – sounds you can just do `readonly: true` on the association declarations in the model.

platform-hub-api/app/models/kubernetes_token.rb
```diff
@@ -1,18 +1,142 @@
...
+  include Audited
+  audited descriptor_field: :name, associated_field: :tokenable
 
-  def self.from_data(attributes)
-    attributes = attributes.with_indifferent_access
-    params = attributes
-      .extract!(:identity_id, :cluster, :uid, :groups, :expire_privileged_at)
-      .merge!(token: ENCRYPTOR.decrypt(attributes[:token]))
-      .with_indifferent_access
+  attr_readonly :token, :name, :uid, :kind, :cluster_id
```
jits: I guess we should be able to update the `name`? (sorry, I know I probably suggested making it readonly in the first place!)

platform-hub-api/app/models/kubernetes_token.rb
```diff
@@ -1,18 +1,142 @@
...
 
-    new(params)
+  before_save :downcase_name
+
+  belongs_to :tokenable, -> { readonly }, polymorphic: true
```
jits: I think you can do `readonly: true` on the association instead.

platform-hub-api/app/models/kubernetes_token.rb
```diff
@@ -1,18 +1,141 @@
...
+    )
+  end
+
+  protected
+
+  def tokenable_type_set_properly
+    if robot?
+      errors.add(:tokenable_type, "must be `User` for robot token") if tokenable_type != 'User'
+    elsif user?
+      errors.add(:tokenable_type, "must be `Idendity` for user token") if tokenable_type != 'Identity'
```
jits: Tyop

platform-hub-api/app/models/kubernetes_token.rb
```diff
@@ -1,18 +1,141 @@
...
+    return unless user?
+    if new_record? && user_token_for_cluster_exists?(cluster)
+      errors.add(:user, "can have only one user token per cluster")
+    end
+  end
+
+  def robot_name_unique_for_given_cluster
+    # For robot token we allow multiple per cluster unique by name
+    return unless robot?
+    if new_record? && robot_token_for_cluster_and_name_exists?(cluster, name)
```
jits: If we do make the `name` editable, then this will need updating (so it also runs on updates).

platform-hub-api/app/models/identity.rb
```diff
@@ -10,6 +10,8 @@ class Identity < ApplicationRecord
     kubernetes: 'kubernetes'
   }
 
+  has_many :tokens, -> { where kind: 'user' }, as: :tokenable, class_name: KubernetesToken
```
jits: Something doesn't seem right about having this association in `Identity` directly, given it only applies to one type of identity object – what do you think about removing it from here, and just using the querying/scopes on `KubernetesToken` directly?

platform-hub-api/app/models/kubernetes_token.rb
jits: There was a reason why I didn't use `attr_readonly` in the `Announcement` model (and resorted to checking things in a `before_save` – I think it's because it doesn't raise an error, so updates won't fail, but the change won't get persisted. I'd rather it errors hard (so we can catch bugs earlier). What do you think?

platform-hub-api/app/models/kubernetes_token.rb
```diff
@@ -1,18 +1,141 @@
...
-    new(params)
+  before_save :downcase_name
+
+  belongs_to :tokenable, -> { readonly }, polymorphic: true
+  belongs_to :cluster, -> { readonly }, class_name: KubernetesCluster
+
+  scope :privileged, -> { where.not(:expire_privileged_at => nil) }
+  scope :by_tokenable, ->(tokenable) { 
+    where("tokenable_type=? AND tokenable_id=?", tokenable.class.name, tokenable.to_param) 
+  }
```
jits: Just noticed that you can now do `where(tokenable: obj)` and Rails will automagically work out the required query 🍾

platform-hub-api/app/models/concerns/read_only_associations.rb
marcinc: It's not used anywhere, accidental commit.

platform-hub-api/app/models/kubernetes_token.rb
marcinc: Regarding `name` - I think we should stick with it being immutable. Is there any use case you have in mind for it? 
Re `attr_readonly` - Yes, that's right. It doesn't raise an error on update - it ignores those fields. I quite like that behaviour. The fact that it doesn't throw won't affect functionality in any way. API already filters out all non-modifiable fields and if you do update directly (e.g. from the console) it won't persist those changes. Can make it throw a dramatic error on update if you prefer that though :)

platform-hub-api/app/models/identity.rb
marcinc: I know it's only relevant to kubernetes identity, however having this association simplifies creation of tokens for given user as it automatically sets tokenable object. Can remove and modify other places but I don't think it'll make things dramatically better. Perhaps just a comment here should suffice?

platform-hub-api/app/models/kubernetes_token.rb
jits: Re: `name` – I was thinking of cases where a typo is made, or needing to change a `_` to a `-` (or vice versa) e.g. to make consistent with other names (like I think we've had to do before). Can keep it immutable and if we need to we'll make it modifiable later.

Re: `attr_readonly` – a "dramatic error" would be amazing ;-) My concern is running some commands in a console, or background job, or etc. and not realising that it hasn't been updated. But okay, let's keep as-is for now.

platform-hub-api/app/models/identity.rb
jits: Yeah that's fine. I would say that we should add a validation to ensure only `kubernetes` identities actually ever have tokens, but it's fine as-is for now.

platform-hub-api/app/models/kubernetes_token.rb
marcinc: Nope. It's `-> { readonly }` now.

jits: Ah! Good to know.

platform-hub-api/app/models/kubernetes_token.rb
marcinc: Why? Since `name` and `cluster` don't change this only need to be triggered on creation.

jits: Yeah, my comment was connected to the suggestion before of making `name` modifiable. But if we're keeping it immutable then no need to change this.

platform-hub-web/src/app/shared/hub-api/hub-api.service.js
```diff
@@ -721,32 +730,73 @@ export const hubApiService = function ($rootScope, $http, $q, logger, events, ap
...
+      throw new Error('"tokenId" argument not specified or empty');
     }
 
-    if (_.isNull(cluster) || _.isEmpty(cluster)) {
-      throw new Error('"cluster" argument not specified or empty');
+    return $http
+      .delete(`${apiEndpoint}/kubernetes/tokens/${tokenId}`)
+      .then(response => {
+        // handle 4xx errors which are not picked up by `catch`.
+        if (response.data.error && response.data.error.status.toString().match(/4../)) {
```
jits: Just realised that this should be `if (response.status.toString().match(/4../))`, in case of 4xx errors that don't contain a body.

PR #138, opened 2017-10-12
Hub admins can now allocate Kubernetes groups to projects or project services

jits:
Note: this doesn't yet include a way to list and modify these allocations. Coming soon!

### DIFF
--- a/platform-hub-api/spec/factories/allocations.rb
+++ b/platform-hub-api/spec/factories/allocations.rb
@@ -1,0 +1,5 @@
+FactoryGirl.define do
+  factory :allocation do
+    
+  end
+end
--- a/platform-hub-web/src/app/projects/services/project-service-selector-popup.html
+++ b/platform-hub-web/src/app/projects/services/project-service-selector-popup.html
@@ -1,0 +1,81 @@
+<md-dialog aria-label="Choose a project service">
+  <md-toolbar class="md-toolbar-tools md-accent">
+    <h2 ng-if="$ctrl.serviceIsOptional">
+      Choose a project or a service from a project
+    </h2>
+    <h2 ng-if="!$ctrl.serviceIsOptional">
+      Choose a service from a project
+    </h2>
+    <span flex></span>
+    <md-button
+      class="close-button md-icon-button"
+      ng-click="$ctrl.cancel()"
+      aria-label="Close dialog">
+      <md-icon>clear</md-icon>
+    </md-button>
+  </md-toolbar>
+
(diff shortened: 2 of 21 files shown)


platform-hub-api/spec/factories/allocations.rb
```diff
@@ -0,0 +1,5 @@
+FactoryGirl.define do
+  factory :allocation do
+
```
marcinc: This factory need a body.

platform-hub-web/src/app/projects/services/project-service-selector-popup.html
```diff
@@ -0,0 +1,79 @@
...
+            ({{p.name}})
+          </md-option>
+        </md-select>
+      </md-input-container>
+
+      <md-input-container class="md-block">
+        <label for="service">Service:</label>
+        <md-select
+          name="service"
+          ng-model="$ctrl.selectedService"
```
marcinc: I think this should also be `required` depending on `serviceIsOptional`.

platform-hub-web/src/app/shared/hub-api/hub-api.service.js
```diff
@@ -682,6 +683,28 @@ export const hubApiService = function ($rootScope, $http, $q, logger, events, ap
     return msg;
   }
 
+  function allocateKubernetesGroup(groupId, projectId, serviceId) {
```
marcinc: Don't you need a default `null` for serviceId?

platform-hub-api/spec/factories/allocations.rb
jits: I suppose I could provide default items for the two different polymorphic associations, but I wanted to leave this up to the consumer of the factory. In which case there's nothing left to put in the body (but still need a factory).

platform-hub-web/src/app/projects/services/project-service-selector-popup.html
jits: I could add that in, but the form level validation is not used in this component – instead we explicitly check the `selectedService` field to enable/disable button.

Mind you, it would provide a nicer UI (highlighted in red if not filled in), so I'll add that.

platform-hub-web/src/app/shared/hub-api/hub-api.service.js
jits: Not necessarily, as it will be `undefined` if not passed in (such dynamic JavaScript ;-).

PR #139, opened 2017-10-16
More work on groups management (mainly: admins can view + delete)

jits:

### DIFF
--- a/platform-hub-api/app/models/allocation.rb
+++ b/platform-hub-api/app/models/allocation.rb
@@ -12,4 +12,17 @@ class Allocation < ApplicationRecord
   validates :allocation_receivable_type, presence: true
   validates :allocation_receivable_id, presence: true
 
+  validate :ensure_uniqueness
+
+  private
+
+  def ensure_uniqueness
+    if Allocation.where(
+      allocatable: self.allocatable,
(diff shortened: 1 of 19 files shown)


platform-hub-api/app/models/allocation.rb
```diff
@@ -12,4 +12,19 @@ class Allocation < ApplicationRecord
...
+  validate :ensure_uniqueness
+
+  private
+
+  def ensure_uniqueness
+    if Allocation.where(
+      allocatable_type: self.allocatable_type,
+      allocatable_id: self.allocatable_id,
+      allocation_receivable_type: self.allocation_receivable_type,
+      allocation_receivable_id: self.allocation_receivable_id
```
marcinc: I guess you can just do this instead? 
```ruby
Allocation.where(
  allocatable: self.allocatable,
  allocation_receivable: self.allocation_receivable
).exists?
```

jits: I think the way I've done it is slightly more resilient, in that if `allocatable` or `allocation_receivable` don't actually exist in the db then you can validate check still continues (rather than erroring hard). Not sure what's the best way here tbh…

PR #140, opened 2017-10-17
Upgrade a bunch of dependencies/libraries

jits:
**IMPORTANT: do not merge into `master` until we've QA'ed this**

I figured that with all the big changes going into the `projects-self-serve-tokens-integration` branch we'd do some proper QA'ing on the `dev` environment before rolling it out to `prod`. So decided to take this opportunity to upgrade some dependencies :)


jits:
> Though maybe worth also upgrading to 5.1.4?

Discussed offline – this is a bit more involved due to potential breaking changes, so we can schedule that in separately.

PR #142, opened 2017-10-19
Refactor the way we do project manager checks

jits:
- Added a new endpoint for checking whether the logged in user has a specified role in a project
- Used this in the project detail page to determine whether to set the `isProjectManager` flag on that page
- Also changed the flow around in the projct detail page to first check admin status and then only continue if that was successful (this way if the API is down, or something is not quite right, we fail early)

### DIFF
--- a/platform-hub-api/spec/controllers/projects_controller_spec.rb
+++ b/platform-hub-api/spec/controllers/projects_controller_spec.rb
@@ -491,6 +491,93 @@ RSpec.describe ProjectsController, type: :controller do
     end
   end
 
+  describe 'GET #role_check' do
+    let :role do
+      'manager'
+    end
+
+    before do
+      @project = create :project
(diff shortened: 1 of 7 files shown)


platform-hub-api/spec/controllers/projects_controller_spec.rb
```diff
@@ -491,6 +491,93 @@
     end
   end
 
+  describe 'GET #role_check' do
+    let :role do
+      'manager'
```
marcinc: worth adding extra spec for case where passed role is not `manager`?

jits: Good idea, though that should be covered by the routing spec (that makes sure it doesn't even make it to the controller if an incorrect role is passed in).

PR #146, opened 2017-10-30
Backend: model validations for checking cluster and groups for kube robot tokens

jits:
This ensures that only allocated clusters and RBAC groups can be used when creating kubernetes tokens of kind `robot`.

---

**Note:** I've gone for a slightly less performant implementation in order to (hopefully) provide a more readable and better structured solution to the different validations. Given that a) tokens won't be created/updated that often, and b) we assume that at most there only be a few groups assigned to a token, this implementation should be "okay" (until proven otherwise :)).

### DIFF
--- a/platform-hub-api/app/models/kubernetes_token.rb
+++ b/platform-hub-api/app/models/kubernetes_token.rb
@@ -40,6 +40,9 @@ class KubernetesToken < ApplicationRecord
   validate :token_must_be_of_expected_length
   validate :one_user_token_per_cluster
   validate :robot_name_unique_for_given_cluster
+  validate :group_names_exist
+  validate :allowed_clusters_only
+  validate :allowed_groups_only
 
   def token=(val)
     self['token'] = ENCRYPTOR.encrypt(val)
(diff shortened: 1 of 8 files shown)


platform-hub-api/app/models/kubernetes_token.rb
```diff
@@ -124,6 +127,54 @@ def robot_name_unique_for_given_cluster
...
+  def allowed_groups_only
+    return unless tokenable.present? && cluster.present?
+
+    # We assume at this point that any group names set actually do exist in the
+    # db (i.e. we expect a previous validation to take care of this)
+    if robot? && groups.present?
+      groups.each do |name|
+        g = KubernetesGroup.where(name: name).first
+
+        allowed = !g.user? &&
```
marcinc: `g.robot? && ...` instead?

platform-hub-api/spec/controllers/kubernetes/tokens_controller_spec.rb
```diff
@@ -199,7 +201,7 @@
       {
         kind: kind,
         cluster_name: @cluster.name,
-        groups: 'group1,group2,group3',
```
marcinc: `groups` come as a string from the UI when creating a token so would probably make sense to leave it like that.

platform-hub-api/app/models/kubernetes_token.rb
```diff
@@ -124,6 +127,54 @@ def robot_name_unique_for_given_cluster
     end
   end
 
+  def groups_exist
+    Array(self.groups).each do |g|
```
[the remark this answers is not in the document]
jits: This is just in case it's `nil`.

platform-hub-api/app/models/kubernetes_token.rb
```diff
@@ -124,6 +127,54 @@ def robot_name_unique_for_given_cluster
     end
   end
 
+  def groups_exist
```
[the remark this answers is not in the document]
jits: Agreed the name could be better – I'll do something like `group_names_exist` to clarify that we're checking that the group names used actually exist.

Re: using `:if` – the line below caters for `nil` and empty lists, so should be ok.

platform-hub-api/app/models/kubernetes_token.rb
jits: Good idea.

platform-hub-api/spec/controllers/kubernetes/tokens_controller_spec.rb
jits: It can be provided as either a string or array – (e.g. for robot tokens it's an Array). And note that with upcoming changes it'll also be an array for user tokens too. We test out the string/array handling in the model tests, and also in `PUT update` test later. So I made this one an array just to make it read slightly better.

PR #148, opened 2017-11-02
Token reveal

marcinc:


platform-hub-api/app/models/kubernetes_token.rb
```diff
@@ -64,6 +64,11 @@ def decrypted_token
     ENCRYPTOR.decrypt(token)
   end
 
+  def obfuscated_token
+    val = decrypted_token
+    val[0..30].gsub(/\w/, 'X') + val[31..35]
```
jits: Slightly faster (possibly):

```ruby
('X' * 30) + val[31..35]
```

marcinc: This one gets rid of dashes so leaving as is :)

platform-hub-web/src/app/kubernetes/kubernetes-robot-tokens-list.html
```diff
@@ -67,7 +67,7 @@
           <strong>Cluster:</strong> {{t.cluster.name}}
         </p>
         <p class="md-body-1">
-          <strong>Token:</strong> {{t.token}}
+          <strong>Token:</strong> <kubernetes-token item="t"/>
```
jits: Deprecated technology alert 😛  – need:

```
<kubernetes-token item="t"></kubernetes-token>
```

(I think the final build will fail otherwise, in certain cases).

PR #149, opened 2017-11-03
Initial backend and API for self serve Kubernetes user tokens for projects

jits:


platform-hub-api/app/models/kubernetes_token.rb
```diff
@@ -69,23 +71,25 @@ def obfuscated_token
...
     expire_privileged_at.present?
   end
 
-  def escalate(privileged_group, expire_in_secs = 600)
-    update_attributes(
+  def escalate(privileged_group_name, expire_in_secs = 600)
+    group = KubernetesGroup.where(name: privileged_group_name).first
+
+    if group.blank?
+      raise ActiveRecord::RecordNotFound
```
marcinc: if you use `.find_by!` above you can skip this check.

platform-hub-api/app/models/kubernetes_token.rb
```diff
@@ -69,23 +71,25 @@ def obfuscated_token
...
+    end
+
+    unless is_group_valid?(group, allow_privileged_groups: true)
+      raise ArgumentError, "specified privileged group '#{privileged_group_name}' is not a valid group for this token (it may not be allocated yet)"
+    end
+
+    # Update the db directly, bypassing validations
```
marcinc: Why do we want to skip validations here?

platform-hub-api/app/models/kubernetes_token.rb
jits: Good idea!

platform-hub-api/app/models/kubernetes_token.rb
jits: The issue here is the `allowed_groups_only` validation check – currently, it won't allow privileged groups to be set (to prevent project admins from explicitly setting this when creating/editing tokens from the UI/API). So this method fails if we run this validation (because it's trying to set a privileged group).

marcinc: I see. Maybe worth putting some extra context in the comment describing why we chose to do that.

jits: Good idea.

PR #153, opened 2017-11-09
New feature flags for tokens sync + tokens escalation of privileges

jits:
This provides finer grained control over the Kubernetes tokens features.

Note:
- The existing `kubernetes_tokens` flag still exists and is used as a "master" toggle of all Kubernetes tokens features in the hub.
- Thus, this commit adds support for configuring and inspecting multiple feature flags for UI routes, shell nav and API endpoints.
- The tokens privilege expirer job will run regardless of the new `kubernetes_tokens_escalate_privilege` flag (but can be turned off using the existing `kubernetes_tokens` flag).

### DIFF
--- a/platform-hub-api/app/jobs/tokens_sync_job.rb
+++ b/platform-hub-api/app/jobs/tokens_sync_job.rb
@@ -6,7 +6,10 @@ class TokensSyncJob < ApplicationJob
   end
 
   def perform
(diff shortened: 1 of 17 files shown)


jits:
@marcinc – 

> Need migration for new flag(s)?

We probably don't – if the flag doesn't exist yet then essentially that feature is turned off, which is what we want when we roll the latest work out. (Then, when you toggle the flag in the UI, the feature gets set properly).

PR #154, opened 2017-11-10
Only return certain fields in the /app_settings API when not logged in

jits:
Only the platform name and platform overview text should be made public to non-logged in users.

### DIFF
--- a/platform-hub-api/app/controllers/app_settings_controller.rb
+++ b/platform-hub-api/app/controllers/app_settings_controller.rb
@@ -1,5 +1,10 @@
 class AppSettingsController < ApiJsonController
 
+  PUBLIC_FIELDS = [
+    'platformName',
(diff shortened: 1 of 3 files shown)


platform-hub-api/app/controllers/app_settings_controller.rb
```diff
@@ -1,5 +1,10 @@
 class AppSettingsController < ApiJsonController
 
+  PUBLIC_FIELDS = [
+    'platformName',
```
marcinc: `platform_name` ?

jits: Unfortunately we've been using `platformName` so far. This was because the web UI actually controls what goes into the app settings – the API has no schema of what can be in here (at the moment). So we've ended up with camel casing here.

PR #164, opened 2017-11-21
Kubernetes Cluster additional properties for generating kubecfg

KashifSaadat:
Adds the following additional fields to Kubernetes Cluster:
- `api_url`: String value containing the full API url for the Kubernetes Cluster, which a user would specify within their kubeconfig (e.g. `https://api.production.example.com:8443`)
- `ca_cert_encoded`: String value of the CA Certificate, base64 encoded

These values can later be used in generating an example kubeconfig for a token assigned to a user, easing the user journey in getting their local environment correctly configured.

### DIFF
--- a/platform-hub-api/db/migrate/20171121125850_add_kubecfg_extra_fields_to_kubernetes_clusters.rb
+++ b/platform-hub-api/db/migrate/20171121125850_add_kubecfg_extra_fields_to_kubernetes_clusters.rb
@@ -1,0 +1,6 @@
+class AddKubecfgExtraFieldsToKubernetesClusters < ActiveRecord::Migration[5.0]
+  def change
+    add_column :kubernetes_clusters, :api_url, :string, null: false
+    add_column :kubernetes_clusters, :ca_cert_encoded, :string, null: false
+  end
+end
--- a/platform-hub-api/db/structure.sql
+++ b/platform-hub-api/db/structure.sql
@@ -3,7 +3,7 @@
 --
 
 -- Dumped from database version 9.6.1
--- Dumped by pg_dump version 9.6.2
+-- Dumped by pg_dump version 10.1
 
 SET statement_timeout = 0;
 SET lock_timeout = 0;
@@ -265,7 +265,9 @@ CREATE TABLE kubernetes_clusters (
     s3_secret_access_key character varying NOT NULL,
     s3_object_key character varying NOT NULL,
     created_at timestamp without time zone NOT NULL,
-    updated_at timestamp without time zone NOT NULL
+    updated_at timestamp without time zone NOT NULL,
+    api_url character varying NOT NULL,
+    ca_cert_encoded character varying NOT NULL
 );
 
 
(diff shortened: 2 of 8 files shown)


jits:
It would be good if this PR also included the actual UI bits to show this to the end user. I say this because: whilst it is useful to set this data, it'd be more compelling (and a better way of testing out the new fields) if this was used for an end-user use case.

platform-hub-api/db/migrate/20171121125850_add_kubecfg_extra_fields_to_kubernetes_clusters.rb
```diff
@@ -0,0 +1,6 @@
+class AddKubecfgExtraFieldsToKubernetesClusters < ActiveRecord::Migration[5.0]
+  def change
+    add_column :kubernetes_clusters, :api_url, :string, null: false
+    add_column :kubernetes_clusters, :ca_cert_encoded, :string, null: false
```
jits: I _think_ this migration will fail for existing data, as existing clusters won't have these fields set, and the they are required to be non-null.

platform-hub-api/db/structure.sql
```diff
@@ -3,7 +3,7 @@
 --
 
 -- Dumped from database version 9.6.1
--- Dumped by pg_dump version 9.6.2
+-- Dumped by pg_dump version 10.1
```
jits: uh oh :D whilst it doesn't like like any pg v10 features have impacted this PR, you should probably use the same version of postgres :)

platform-hub-api/db/structure.sql
```diff
@@ -265,7 +265,9 @@ CREATE TABLE kubernetes_clusters (
     s3_secret_access_key character varying NOT NULL,
     s3_object_key character varying NOT NULL,
     created_at timestamp without time zone NOT NULL,
-    updated_at timestamp without time zone NOT NULL
+    updated_at timestamp without time zone NOT NULL,
+    api_url character varying NOT NULL,
```
marcinc: inject two new fields before create_at pls.

jits: This is auto-generated

platform-hub-api/db/migrate/20171121125850_add_kubecfg_extra_fields_to_kubernetes_clusters.rb
KashifSaadat: Yeah, I was thinking to make them optional and then we don't show the generate kubecfg button if the values are null. What do you think? Or should we somehow fill existing records with dummy data / set empty strings?

platform-hub-api/db/structure.sql
marcinc: :) yeah, missed that completely...

platform-hub-api/db/structure.sql
KashifSaadat: whoops, that would have been when I ran `bin/rails db:migrate`. I'll downgrade my local pg version.

jits:
Closing this particular PR as it's a little behind on `master` – instead of resolving the merge conflicts it'll be easier to redo the data migration etc. in a separate branch.

With credit to @KashifSaadat for instigating this! :)

PR #165, opened 2017-11-21
Handle deletions of projects, services, allocations, groups, clusters and tokens

jits:
* [x] For deletion of projects
    * [x] Delete all services
        * [x] Make sure this cascades down to robot tokens too!
    * [x] Delete all user tokens for that project
    * [x] Delete all allocations (with clusters)
        * [x] Make sure this cascades down (see later)
* [x] For deletion of services
    * [x] Delete all robot tokens for that service
    * [x] Delete all allocations
        * [x] Make sure this cascades down (see later)
* [x] For deletion of allocations
    * [x] If allocatable_type == 'KubernetesCluster' -> delete all tokens for allocation_receivable project that are for that cluster
    * [x] If allocatable_type == 'KubernetesGroup` -> make sure to remove that group from all tokens that have it, for tokens that belong to either the allocation_receivable project (user tokens or robot tokens) or allocation_receivable service of (robot tokens only for service, or user token)
* [x] For deletion of groups
    * [x] Remove that particular group from all tokens that have it
* [x] For deletion of a cluster
    * [x] Delete all tokens
    * [x] Remove the cluster from any KubernetesGroup#restricted_to_clusters list
        * [x] If group has just the one cluster in `restricted_to_clusters` then we actually need to delete the group (since otherwise `restricted_to_clusters` would become empty and allow all clusters)
* [x] For deletion of a token
    * [x] Rejoice!

### DIFF
--- a/platform-hub-api/spec/factories/kubernetes_tokens.rb
+++ b/platform-hub-api/spec/factories/kubernetes_tokens.rb
@@ -43,7 +43,7 @@ FactoryGirl.define do
 
       trait :user_is_not_member_of_project do
         after(:build) do |token, _|
-          ProjectMembership.where(project_id: token.project_id, user_id: token.tokenable.user_id).delete_all
+          ProjectMembership.where(project_id: token.project_id, user_id: token.tokenable.user_id).map(&:destroy)
         end
       end
     end
--- a/platform-hub-api/spec/factories/kubernetes_tokens.rb
+++ b/platform-hub-api/spec/factories/kubernetes_tokens.rb
@@ -12,10 +12,10 @@ FactoryGirl.define do
       name { "user_#{SecureRandom.uuid}@example.com" }
 
       transient do
-        groups_count 2
+        groups_count 0
       end
 
-      after(:build) do |token, _|
+      after(:build) do |token, evaluator|
         if token.cluster.blank? && token.project.present?
           token.cluster = create :kubernetes_cluster, allocate_to: token.project
         end
@@ -24,10 +24,14 @@ FactoryGirl.define do
           create :project_membership, project: token.project, user: token.tokenable.user
         end
 
-        if token.groups.nil? && !evaluator.groups_count.zero?
-          token.groups = evaluator.groups_count.map do |i|
-            create :kubernetes_group, :not_privileged, :for_user, allocate_to: token.project
-          end
+        if token.groups.blank? && !evaluator.groups_count.zero?
+          token.groups = create_list(
+            :kubernetes_group,
+            evaluator.groups_count,
+            :not_privileged,
+            :for_user,
+            allocate_to: token.project
+          ).map(&:name)
         end
       end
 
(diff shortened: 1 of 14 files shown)


KashifSaadat:
Group deletion: We may need to test what the behaviour is when escalating to use a privileged group, and then deleting that group. Does it leave any weird config behind?

Cluster deletion:
- Can trigger group deletion
- Group allocations to a project/service should be correctly removed
- Cluster allocation should be correctly removed from a project

In all cases it would be helpful to provide a confirmation dialog, and a log of all the resources that will be affected (including cascading deletions / modifications).

jits:
@KashifSaadat –

> Group deletion: We may need to test what the behaviour is when escalating to use a privileged group, and then deleting that group. Does it leave any weird config behind?

Good question – in theory it should just remove that group (though the token will still be in privileged 'mode' until it expires).

> Cluster deletion: […]

Nice one! Will add more logic and/or test cases for these.

> In all cases it would be helpful to provide a confirmation dialog, and a log of all the resources that will be affected (including cascading deletions / modifications).

Quite difficult to do right :) So v2++

Note: all deletions do have a confirmation dialog in the app, just that it won't be easy to then say what exactly will be deleted (without some significant work).

platform-hub-api/spec/factories/kubernetes_tokens.rb
```diff
@@ -39,7 +43,7 @@
 
       trait :user_is_not_member_of_project do
         after(:build) do |token, _|
-          ProjectMembership.where(project_id: token.project_id, user_id: token.tokenable.user_id).delete_all
+          ProjectMembership.where(project_id: token.project_id, user_id: token.tokenable.user_id).map(&:destroy)
```
marcinc: or just 
```ruby
ProjectMembership.destroy_all(project_id: token.project_id, user_id: token.tokenable.user_id)
```

platform-hub-api/spec/integration/tokens_deletion_integration_spec.rb
```diff
@@ -0,0 +1,292 @@
...
+
+      deleted KubernetesToken, other_user_token.id
+      deleted KubernetesToken, other_robot_token.id
+
+      exists KubernetesToken, other_other_user_token.id
+      exists KubernetesToken, other_other_robot_token.id
+    end
+  end
+
+  describe 'on project membership deletion' do
```
marcinc: `on user's project membership deletion` to make it clearer?

platform-hub-api/spec/factories/kubernetes_tokens.rb
jits: I was under the impression that `destroy_all` would not execute any `after_destroy` callbacks on the record being deleted.

platform-hub-api/spec/integration/tokens_deletion_integration_spec.rb
jits: Good idea, will update.

platform-hub-api/spec/factories/kubernetes_tokens.rb
marcinc: I think it does but it's a minor thing. Can leave as is.

PR #167, opened 2017-11-27
Various UI and backend improvements from user feedback

jits:
Mainly for the projects self serve tokens feature.

### DIFF
--- a/platform-hub-api/app/models/kubernetes_group.rb
+++ b/platform-hub-api/app/models/kubernetes_group.rb
@@ -38,6 +38,8 @@ class KubernetesGroup < ApplicationRecord
   }
   validates :target, presence: true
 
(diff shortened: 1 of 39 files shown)


platform-hub-api/app/models/kubernetes_group.rb
```diff
@@ -74,4 +76,12 @@ def handle_destroy
     KubernetesToken.update_all_group_removal self.name
   end
 
+  def cluster_names_exists
+    Array(self.restricted_to_clusters).each do |name|
```
marcinc: Do you need to cast to Array specifically?

jits: I suppose it's not _necessary_; it's more to be defensive, in case it's `nil`.

PR #169, opened 2017-11-29
Hub admins can now manage Kubernetes namespaces, assigned to project services

jits:
Note: there's currently no relationship/interaction between RBAC groups and namespaces within the hub – we can solve this in a later PR (depending on _if_ and _how_ we might want to do that).

### DIFF
--- a/platform-hub-api/app/models/kubernetes_namespace.rb
+++ b/platform-hub-api/app/models/kubernetes_namespace.rb
@@ -1,0 +1,40 @@
+class KubernetesNamespace < ApplicationRecord
+
+  NAME_REGEX = /\A[a-z0-9]([-a-z0-9]*[a-z0-9])?\z/
+
+  include Audited
+
+  audited descriptor_field: :name, associated_field: :service
+
+  belongs_to :service, -> { readonly }
+  belongs_to :cluster, -> { readonly }, class_name: KubernetesCluster
+
+  scope :by_service, ->(s) { where(service: s) }
+  scope :by_cluster, ->(c) { where(cluster: c) }
+
+  validates :name,
+    format: {
+      with: NAME_REGEX,
+      message: "should consist of lowercase letters, numbers and dashes, and must start and end with an alphanumeric character"
+    },
+    uniqueness: {
+      scope: :cluster_id,
+      message: "already exists for the cluster"
+    }
+
+  validate :allowed_clusters_only
+
+  protected
+
+  def allowed_clusters_only
+    return unless service.present? && cluster.present?
+
+    unless Allocation.exists?(
+      allocatable: cluster,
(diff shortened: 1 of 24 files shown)


platform-hub-api/app/models/kubernetes_namespace.rb
```diff
@@ -0,0 +1,40 @@
...
+
+  validates :name,
+    format: {
+      with: NAME_REGEX,
+      message: "should consist of lowercase letters, numbers and dashes, and must start and end with an alphanumeric character"
+    },
+    uniqueness: {
+      scope: :cluster_id,
+      message: "already exists for the cluster"
+    }
```
marcinc: Should we also validate description presence?

platform-hub-web/src/app/kubernetes/kubernetes-namespaces-form.component.js
```diff
@@ -0,0 +1,188 @@
...
+    ctrl.loading = true;
+
+    ctrl.isNew = !namespaceId;
+    ctrl.namespace = null;
+
+    if (
+      (!fromProject && fromService) ||
+      (fromProject && !fromService)
+    ) {
+      logger.error('Bug detected: both `fromProject` and `fromService` need to be provided, or neither');
```
marcinc: nice error message :) 🐛

platform-hub-web/src/app/kubernetes/kubernetes-namespaces-form.html
```diff
@@ -0,0 +1,156 @@
...
+              required
+              ng-pattern="/^[a-z0-9]([-a-z0-9]*[a-z0-9])?$/">
+            <div ng-messages="$ctrl.kubernetesNamespaceForm.name.$error">
+              <div ng-message="required">This is required.</div>
+              <div ng-message="pattern">Only lowercase letters, numbers and dashes, and must start and end with an alphanumeric character.</div>
+            </div>
+          </md-input-container>
+
+          <md-input-container class="md-block">
+            <label for="description">Description</label>
```
marcinc: `required` if we decide this should be always present.

platform-hub-web/src/app/kubernetes/kubernetes-namespaces-form.component.js
jits: 👍  bit defensive coding here; it should never happen… but you never know ;-)

platform-hub-api/app/models/kubernetes_namespace.rb
jits: I was thinking we make description optional for namespaces. Seems like the namespace name alone is often descriptive enough, so requiring a description would just make people put the same name as the description.

PR #171, opened 2017-12-13
Initial work on generating and showing costs reports

jits:
~~Apologies for the big commit – it's a difficult one to split up, as it's mostly new and interdependent stuff.~~ I've now split this into some separate commits :)

Note: this is just the initial work on this feature; expect more PRs later to finish it off.


platform-hub-api/app/models/kubernetes_cluster.rb
```diff
@@ -26,6 +27,13 @@ class KubernetesCluster < ApplicationRecord
   validates :description, :s3_region, :s3_bucket_name, :s3_object_key,
             :s3_access_key_id, :s3_secret_access_key, presence: true
 
+  validates :aws_account_id,
```
marcinc: shouldn't this be a required field?

platform-hub-api/app/services/project_bills_query_service.rb
```diff
@@ -0,0 +1,24 @@
+module ProjectBillsQueryService
+  extend self
+
+  def fetch project_id
+    # Don't allow } and ' as they can be used for SQL injection in this particular case
+    raise ArgumentError if project_id.count("}'") > 0
```
marcinc: Could potentially sanitize query below to automatically take care of it?

platform-hub-api/app/controllers/costs_reports_controller.rb
```diff
@@ -0,0 +1,146 @@
...
+      :year,
+      :month,
+      :notes,
+      :billing_file,
+      :metrics_file
+    )
+
+    # Below is a workaround until Rails 5.1 lands and we can use the `foo: [{}]` syntax to permit the whole array of hashes
+    allowed_params[:config] =
+    params[:costs_report][:config]
```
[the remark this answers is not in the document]
jits: Ah yes, not sure why that's on two lines – will change.

platform-hub-api/app/models/kubernetes_cluster.rb
jits: Not necessarily; also, can't make it required as it will affect existing records.

marcinc: Sure. Just thought it's something we know upfront usually so could make it a required field.

platform-hub-api/app/services/costs_report_generator_service.rb
```diff
@@ -0,0 +1,454 @@
...
+      else
+        non_shared_accounts_by_id[account_id] = a
+      end
+    end
+
+    namespaces_by_account_and_name = namespaces.each_with_object(HashInitializer[:hash]) do |n, obj|
+      obj[n[:account_id]][n[:namespace_name]] = n
+    end
+
+    bills_per_day_by_account = billing_data[2..-1].each_with_object(HashInitializer[:hash]) do |r, obj|
```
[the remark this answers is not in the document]
jits: Initially, I did "pre-process" the CSV data in to arrays of hashes, etc. But seemed overkill at the time. The issue is (and this relates to your question re: test coverage) we're not yet sure how this code will evolve – it will certainly change a fair bit when we do the "v2" costs (of layering in AWS resource costs, etc.) and I'm keen on keeping everything isolated to this one method/file for now (even if it is a tad messy). Once this matures we'll have a better idea of how to de-compose it into smaller and more understandable chunks.

platform-hub-api/app/services/costs_report_generator_service.rb
```diff
@@ -0,0 +1,454 @@
...
+            }
+          end
+
+          day_cluster_bills = account_bills.each_with_object({}, &account_to_clusters_proc)
+
+          service_hash['daily'][day] = {
+            'clusters' => day_cluster_bills,
+            'shared_services' => shared_bills_per_day_by_project_and_service[day][project_id][service_id].round(2)
+          }
+
```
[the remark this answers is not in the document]
jits: +1 I expect we definitely will once this matures. This method is the anti-thesis of how code "should be done", but it serves a purpose of getting us to a v1 quicker (I spent some time thinking through breaking this up, but it distracted from actually getting it working). If we had another sprint to work on this I possible would have taken that next step. For now, I think we should address this in the upcoming "v2" implementation.

platform-hub-api/app/services/filestore_service.rb
```diff
@@ -0,0 +1,51 @@
...
+  def get name
+    key = generate_key name
+    @bucket.object(key).get().body.read
+  end
+
+  private
+
+  def generate_key name
+    if @object_prefix.present?
+      "#{@object_prefix}/#{name}"
```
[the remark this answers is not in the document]
jits: I'm not sure if it will?

platform-hub-api/app/services/project_bills_query_service.rb
jits: Can't use regular SQL quote sanitisation as the below query is a special Postgres syntax (sanitisation breaks it).

platform-hub-api/app/models/kubernetes_cluster.rb
jits: We _could_ make it required from now on, but…
a) We wouldn't be able to add a non-NULL db constraint (which is not ideal), due to existing records (unless we defaulted to some value).
b) Even if we didn't do (a) we could still make it so when you next update a cluster you have to provide the account ID… but I can imagine a situation where someone is trying to update some other important value (like the S3 bucket details) and it becomes frustrating because they have to dig up the AWS account ID in that moment,.

marcinc: OK, we can leave as is and take care of it later. I guess accounts without ID won't be taken into account for costing so that's OK for now.

platform-hub-api/app/services/costs_report_generator_service.rb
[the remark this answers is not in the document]
marcinc: Cool cool. Let's refactor / test that in future iterations.

platform-hub-api/app/services/costs_report_generator_service.rb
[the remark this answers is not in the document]
marcinc: 👍

platform-hub-api/app/services/filestore_service.rb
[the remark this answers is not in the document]
marcinc: I guess it'll be fine.

platform-hub-api/app/services/project_bills_query_service.rb
marcinc: Ah! all good then.

PR #173, opened 2017-12-18
Add an extra publish step for Costs Reports

jits:
… before they get shown to projects.

Note: once published, a report cannot be edited, nor recreated via the reports form (though you can still explicitly delete it and recreate it in two separate steps, as a special case capability).

### DIFF
--- a/platform-hub-api/spec/models/costs_report_spec.rb
+++ b/platform-hub-api/spec/models/costs_report_spec.rb
@@ -18,6 +18,34 @@ RSpec.describe CostsReport, type: :model do
     end
   end
 
+  describe '.already_published?' do
+    context 'when it doesn\'t exist' do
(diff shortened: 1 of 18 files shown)


platform-hub-api/spec/models/costs_report_spec.rb
```diff
@@ -18,6 +18,34 @@
...
+        end
+      end
+
+      context 'after being published' do
+        before do
+          subject.publish!
+        end
+
+        it 'should return false' do
+          expect(CostsReport.already_published?(2017, 'Jan')).to be false
```
marcinc: I guess this should assert true?

jits: Heh! That's a good catch :)

PR #176, opened 2017-12-28
Helper to set up Kube config for a user

jits:


platform-hub-api/spec/controllers/kubernetes/clusters_controller_spec.rb
```diff
@@ -218,6 +224,8 @@
           expect(updated.aws_account_id).to eq put_data[:aws_account_id].to_i
           expect(updated.s3_region).to eq put_data[:s3_region]
           expect(updated.s3_bucket_name).to eq @cluster.s3_bucket_name
+          expect(updated.api_url).to eq put_data[:api_url]
+          expect(updated.ca_cert_encoded).to eq put_data[:ca_cert_encoded]
```
KashifSaadat: `ca_cert_encoded` is unset in the `put_data` above, is this valid?

platform-hub-web/src/app/shared/kube-config-helper-popup.controller.js
```diff
@@ -0,0 +1,22 @@
...
+  const ctrl = this;
+
+  ctrl.copyIcon = icons.copy;
+
+  ctrl.clusterName = token.cluster.name;
+
+  const contextName = `${token.cluster.name}_${token.project.shortname}`;
+  const userName = `${kubeId}_${contextName}`;
+
+  ctrl.clusterCommands = `kubectl config set-cluster ${token.cluster.name} --server=${token.cluster.api_url} && kubectl config set clusters.${token.cluster.name}.certificate-authority-data ${token.cluster.ca_cert_encoded}`;
```
KashifSaadat: Do we allow dots in the cluster name definition? If so, it might break the second half of this particular command.

marcinc: Nope. it's only alphanumerics and dashes so should be OK.

platform-hub-api/spec/controllers/kubernetes/clusters_controller_spec.rb
jits: Oh nice catch – this line should be:

```
expect(updated.ca_cert_encoded).to eq @cluster.ca_cert_encoded
```

(note that the PUT operation is additive, in that it won't unset any fields that are not explicitly provided)

platform-hub-web/src/app/shared/kube-config-helper-popup.controller.js
jits: Looks like [we don't allow dots](https://github.com/UKHomeOffice/platform-hub/blob/8c499f4a80a7af05fe112a847ede4206de3a609d/platform-hub-api/app/models/kubernetes_cluster.rb#L3) so should be all good (but I like your line of thinking!)

PR #184, opened 2018-01-11
New 'limited admin' role

jits:
Currently only granted read access to the costs reports listing and details pages.

### DIFF
--- a/platform-hub-api/app/models/ability.rb
+++ b/platform-hub-api/app/models/ability.rb
@@ -6,8 +6,16 @@ class Ability
     # IMPORTANT: to understand the caveats when using blocks to specify
     # abilities, see: https://github.com/CanCanCommunity/cancancan/wiki/Defining-Abilities-with-Blocks
 
-    # This will take precedence over everything below!
+
+    # Hub admin role
+    # Note: this will take precedence over everything below!
     can :manage, :all if user.admin?
+
+
+    #  Special hub limited_admin role
+    if user.limited_admin?
+      can :read, CostsReport
+    end
 
 
     # Projects
--- a/platform-hub-api/spec/rails_helper.rb
+++ b/platform-hub-api/spec/rails_helper.rb
@@ -66,13 +66,9 @@ RSpec.configure do |config|
 
 
   # Support and shared stuff
-  config.include JsonResponseHelper, type: :request
   config.include JsonResponseHelper, type: :controller
   config.include UserActivationHelpers, type: :controller
-  config.include_context 'authentication helpers', type: :request
   config.include_context 'authentication helpers', type: :controller
-  config.include_context 'admin authorization helpers', type: :request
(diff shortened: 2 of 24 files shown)


platform-hub-api/app/models/ability.rb
```diff
@@ -6,10 +6,18 @@ def initialize(user)
     # IMPORTANT: to understand the caveats when using blocks to specify
     # abilities, see: https://github.com/CanCanCommunity/cancancan/wiki/Defining-Abilities-with-Blocks
 
-    # This will take precedence over everything below!
+
+    # Hub admin role
+    # Note: this will take precedence over everything below!
```
KashifSaadat: Is this comment still relevant, as you can only have one of the roles active now?

platform-hub-api/spec/rails_helper.rb
```diff
@@ -66,13 +66,11 @@
...
   # Support and shared stuff
-  config.include JsonResponseHelper, type: :request
   config.include JsonResponseHelper, type: :controller
   config.include UserActivationHelpers, type: :controller
-  config.include_context 'authentication helpers', type: :request
   config.include_context 'authentication helpers', type: :controller
-  config.include_context 'admin authorization helpers', type: :request
-  config.include_context 'admin authorization helpers', type: :controller
+  # config.include_context 'hub admin authorization helpers', type: :controller
+  # config.include_context 'hub limited admin authorization helpers', type: :controller
```
KashifSaadat: These still needed?

platform-hub-web/src/app/app.routes.js
```diff
@@ -771,15 +771,15 @@ export const appRoutes = function ($stateProvider, $urlRouterProvider, $location
...
+          rolesPermitted: ['admin', 'limited_admin']
         }
       })
       .state('costs-reports.new', {
         url: '/new',
         component: CostsReportsForm,
         data: {
           authenticate: true,
-          rolePermitted: 'admin'
+          rolesPermitted: ['admin']
```
KashifSaadat: Would we also want to give `limited_admin` people the ability to create new cost reports?

platform-hub-api/app/models/ability.rb
jits: It's still relevant, but just in the context of how `cancancan` uses this file – i.e. the _ordering_ of these abilities/rules are really important.

platform-hub-api/spec/rails_helper.rb
jits: ah good catch, I need to remove

platform-hub-web/src/app/app.routes.js
jits: I don't believe we'll be doing that at the moment – I think it's purely for consumption at this stage (though it's easy to add access for it later, if needed).

PR #185, opened 2018-01-11
Backend: onboard/offboard to Github when deactivating a user

jits:
Note: this is a "non blocking" operation, in that if onboard/offboard fails for whatever reason then the activation/deactivation of the user still continues.

### DIFF
--- a/platform-hub-api/app/services/user_activation_service.rb
+++ b/platform-hub-api/app/services/user_activation_service.rb
@@ -1,14 +1,33 @@
 module UserActivationService
   extend Agents::KeycloakAgentInstance
+  extend Agents::GitHubAgentInstance
   extend self
 
   def activate! user
     keycloak_agent_service.activate_user(user)
+
+    if user.github_identity
+      begin
+        git_hub_agent_service.onboard_user user
+      rescue => e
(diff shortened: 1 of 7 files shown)


platform-hub-api/app/services/user_activation_service.rb
```diff
@@ -1,14 +1,33 @@
...
   end
 
   def deactivate! user
     keycloak_agent_service.deactivate_user(user)
+
+    if user.github_identity
+      begin
+        git_hub_agent_service.offboard_user user
+      rescue => e
+        logger.error "Failed to offboard from Github whilst deactivating user. Still continuing with deactivation. Exception type: #{e.class.name}. Message: #{e.message}"
```
KashifSaadat: If this does occur when deactivating a user, will the error log be passed up in the response and to the UI? If it does happen, the Hub Admin should ideally be notified clearly so they can take manual action to investigate / off-board as necessary.

jits: No, it currently won't. Right now there's no easy way to propagate this _warning_ up to the admin that carried out the action and show it as a result of the API call. I'd suggest instead of adding this capability, we make it part of the overall offboarding RL process to double check that the user has been offboarded (which we should do anyway; as there could be a number of different reasons why this operation fails).

PR #196, opened 2018-01-31
Web: allow `per_page` override on certain views

jits:
You can now control the number of items returned per page in certain views by adding the `per_page=XX` query parameter to the URL.

The following views support this:
- Users list
- Kube groups list
- Kube namespaces list
- Kube robot tokens list

### DIFF
--- a/platform-hub-web/src/app/shared/hub-api/api-pagination.factory.js
+++ b/platform-hub-web/src/app/shared/hub-api/api-pagination.factory.js
@@ -1,0 +1,55 @@
+/* eslint camelcase: 0, object-shorthand: 0 */
+
+export const apiPagination = function (_) {
+  'ngInject';
+
+  const DEFAULT_PER_PAGE = 10;
+
+  let perPage;
+
+  resetPerPage();
+
+  return {
+    setPerPage,
+    resetPerPage,
+    withPaginationParams,
+    handlePaginatedResponse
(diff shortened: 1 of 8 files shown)


platform-hub-web/src/app/shared/hub-api/api-pagination.factory.js
```diff
@@ -0,0 +1,55 @@
...
+
+  function handlePaginatedResponse(response) {
+    const items = response.data;
+
+    const headers = response.headers();
+
+    if (headers.total && headers['per-page']) {
+      items.pagination = {
+        total: parseInt(headers.total, 10),
+        perPage: parseInt(headers['per-page'], 10)
```
KashifSaadat: In the `app.routes.js` it is `per_page`, should this be an underscore here too?

jits: It's a fair question/point – the _header_ from the API uses `per-page`, but I've made it use `per_page` in the URL for the query param. Personally I prefer using the best form for each particular context, and for query params I tend to prefer underscored names. In JavaScript the variable names are `perPage`, so we have the third form anyway.

Happy to change if there's more strong feelings towards this (though I'd rather not change the header as that's the best form for header names, so it'd be changing the query param [if the router actually allows this form]).

PR #197, opened 2018-02-09
Initial tool for analysing costs and resource usage for a project

jits:
Currently only accessible by hub admins.

This is the first step towards a more elaborate tool for finely analysing a project's costs and resource usage per cluster.

For now this focuses on using the exact same data used for generating costs reports (billing + metrics) to show resource utilisation for each *namespace* of a project.

### DIFF
--- a/platform-hub-api/app/services/costs_and_resources_analysis_service.rb
+++ b/platform-hub-api/app/services/costs_and_resources_analysis_service.rb
@@ -1,0 +1,189 @@
+module CostsAndResourcesAnalysisService
+
+  extend self
+
+  def source_data project, billing_csv_string, metrics_csv_string
+    prepare_results, billing_data, metrics_data = CostsReportGeneratorService.prepare_and_return_data(
+      billing_csv_string,
+      metrics_csv_string
+    )
+
+    mapped_accounts = prepare_results[:accounts][:mapped]
+    mapped_namespaces = prepare_results[:namespaces][:mapped]
+
+    project_namespaces = mapped_namespaces.select { |n| n[:project_id] == project.id }
+
+    calculate_daily_usage(
+      mapped_accounts,
+      project_namespaces,
+      prepare_results[:metrics],
+      billing_data,
+      metrics_data
+    )
+  end
+
+  private
+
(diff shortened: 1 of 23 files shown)


platform-hub-api/app/services/costs_and_resources_analysis_service.rb
```diff
@@ -0,0 +1,193 @@
...
+      project_namespaces,
+      prepare_results[:metrics],
+      billing_data,
+      metrics_data
+    )
+  end
+
+  private
+
+  def calculate_daily_usage accounts, namespaces, metrics, billing_data, metrics_data
```
marcinc: This method is pretty massive... lots of things going on. Do you think it's worth to split it in separate helper methods and just leave the core logic here? Would be much easier to review.

jits: It's a similar situation to the costs report generation code you reviewed previously – I realise it's harder to review and this is atypical code; I could spend more time refactoring and splitting it up, but it's very likely this is all going to be changed / massively redone in upcoming work (especially with the new costs stuff discussed yesterday). And the only refactoring / splitting I can think of to do right now is to break out the different loops – that construct data from other data and assign to intermediate variables – into individual methods, which seems purely cosmetic.

Hopefully the fact that I've documented the expected output format + added comments along the way + named variables descriptively helps with reviewing? If not, we can go through it together and I can walk you through the steps?

marcinc: Fair play. Let's leave as is given it's very likely to change.

jits: Just to elaborate on this a bit more – I had initially started refactoring stuff into modules/classes like `CostsReportBillingDataHelper`, `CostsReportMetricsDataHelper` and `CostsReportDataUtils` but things got complex very quickly with lots of cross communication between the pieces. Mainly because of the nature of the data processing (lots of intermediate data cache structures and mappings) and also because the abstractions are not yet mature and will take some time before they do.

Hopefully in the next big piece of costs work (upcoming) I can get this to a better place :)

PR #199, opened 2018-02-16
Add `aws_region` to `KubernetesCluster` records

jits:
This allows us to support multiple Kubernetes clusters within the same AWS account ID, since we can now differentiate them by region. We'll be using this capability in upcoming costs reports work.


gambol99:
LGTM

jits:
@marcinc –

> Given there is no plans to have more than 1 kube cluster per region.

Yeah exactly – we would need to re-address again if this changes.

PR #200, opened 2018-02-21
Add some uniqueness constraints for AWS account ID and region in Kubernetes clusters

jits:
This ensures we only ever have one KubernetesCluster record per AWS account + region pair.


platform-hub-api/app/models/kubernetes_cluster.rb
```diff
@@ -32,6 +32,25 @@ class KubernetesCluster < ApplicationRecord
...
+    }
+
+  validates :aws_account_id,
+    presence: {
+      message: "can't be blank if AWS region is set"
+    },
+    if: ->(c) { c.aws_region.present? }
+
+  validates :aws_region,
+    allow_nil: true,
```
marcinc: this should be not optional I think. I guess you did it so migration doesn't break existing entries?

jits:
Turns out we can't make this assumption at the moment; instead we'll be adding aliases, so we can map to clusters in the hub. Closing this off.

PR #201, opened 2018-02-22
Support multiple unique aliases for Kubernetes clusters

jits:
---

This PR also includes some improvements – see the individual commits for details.

### DIFF
--- a/platform-hub-api/app/controllers/kubernetes/clusters_controller.rb
+++ b/platform-hub-api/app/controllers/kubernetes/clusters_controller.rb
@@ -116,7 +116,8 @@ class Kubernetes::ClustersController < ApiJsonController
       :s3_secret_access_key,
       :s3_object_key,
       :api_url,
-      :ca_cert_encoded
+      :ca_cert_encoded,
+      aliases: []
     )
   end
 
--- a/platform-hub-api/spec/models/kubernetes_cluster_spec.rb
+++ b/platform-hub-api/spec/models/kubernetes_cluster_spec.rb
@@ -2,6 +2,70 @@ require 'rails_helper'
 
 RSpec.describe KubernetesCluster, type: :model do
 
+  describe 'scopes' do
+
+    describe 'by_alias' do
+      let(:value) { 'foo' }
+
+      subject { KubernetesCluster.by_alias(value) }
+
+      context 'when no aliases are set' do
+        let!(:cluster) { create :kubernetes_cluster, aliases: [] }
+
+        it 'should not find any clusters' do
+          expect(subject.count).to be 0
+        end
+      end
+
+      context 'when the value is set in an aliases list' do
+        let!(:cluster) { create :kubernetes_cluster, aliases: [ value, 'another' ] }
+
+        it 'should find the cluster' do
+          expect(subject.entries).to eq [ cluster ]
+        end
+      end
+
+      context 'when the value is not used' do
+        let!(:cluster) { create :kubernetes_cluster, aliases: [ 'another' ] }
+
+        it 'should not find any clusters' do
+          expect(subject.count).to be 0
+        end
+      end
+    end
+
+    describe 'by_name_or_alias' do
+      let(:value) { 'foo' }
+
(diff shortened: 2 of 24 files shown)


platform-hub-api/app/controllers/kubernetes/clusters_controller.rb
```diff
@@ -116,7 +116,8 @@ def cluster_params
       :s3_secret_access_key,
       :s3_object_key,
       :api_url,
-      :ca_cert_encoded
+      :ca_cert_encoded,
+      aliases: [],
```
KashifSaadat: Drop the comma on the end? Consistent with other definitions

jits: Yeah this is one of those weird things Ruby allows that I'm never consistent with 😄 On the one hand it's useful to keep in since future updates to the list will result in less lines of diff. On the other hand… it doesn't matter as much :) Will update to keep consistent, at least.

platform-hub-api/spec/models/kubernetes_cluster_spec.rb
```diff
@@ -2,6 +2,70 @@
...
+        it 'should not find any clusters' do
+          expect(subject.count).to be 0
+        end
+      end
+    end
+
+    describe 'by_name_or_alias' do
+      let(:value) { 'foo' }
+
+      subject { KubernetesCluster.by_name_or_alias(value) }
```
KashifSaadat: Super nitpick here: Could possibly include an extra test case for `by_name_or_alias` where no name is defined. You have a similar test case for `by_alias` where no name or alias is provided.

jits: I do like your thinking here – only thing is `name` is actually a required field, so will always have a value.

Unless I've misunderstood what you meant?

platform-hub-web/src/app/kubernetes/kubernetes-clusters-form.component.js
```diff
@@ -15,12 +15,16 @@ function KubernetesClustersFormController($state, KubernetesClusters, logger) {
...
+  ctrl.separatorKeys = chipsHelpers.separatorKeys;
+  ctrl.separatorKeysHelpText = chipsHelpers.separatorKeysHelpText;
+
   ctrl.loading = true;
   ctrl.saving = false;
   ctrl.isNew = true;
   ctrl.cluster = null;
 
   ctrl.createOrUpdate = createOrUpdate;
+  ctrl.processAliasesChip = processAliasesChip;
```
KashifSaadat: Is `processAliasesChip` defined somewhere? I can't find it

jits: It should be defined towards the bottom of this file.

platform-hub-api/spec/models/kubernetes_cluster_spec.rb
KashifSaadat: Aha okay that's fine, I was just thrown off by the earlier tests that do: `let!(:cluster) { create :kubernetes_cluster, aliases: [] }` (no name defined)

platform-hub-web/src/app/kubernetes/kubernetes-clusters-form.component.js
KashifSaadat: Oh yes my mistake! 👍

platform-hub-api/spec/models/kubernetes_cluster_spec.rb
jits: Ahhh yeah, we're using "factories" in our tests to help build/create pre-canned objects that have necessary bits already set up (and then you can override as you please). Check out https://github.com/UKHomeOffice/platform-hub/blob/master/platform-hub-api/spec/factories/kubernetes_clusters.rb for the `KubernetesCluster` factory.

The really cool thing about this is you don't need to do lots of object/model initialisation in many different places :)

PR #206, opened 2018-03-14
Add `costs_bucket` to `KubernetesCluster` records

jits:


platform-hub-api/app/controllers/kubernetes/clusters_controller.rb
```diff
@@ -110,6 +110,7 @@ def cluster_params
       :description,
       :aws_account_id,
       :aws_region,
+      :costs_bucket,
```
marcinc: Should we only allow certain values? Or at least make sure that naming convention is right?

jits: Yeah, I was thinking about this: was going to add a separate thing in the app settings (or similar) to have `allowed_costs_buckets` and then make this field limited to those values. It's just a bit of a faff really :) So I wonder if maybe we can get away with not doing this and relying on the hub admin being disciplined enough to add the right values :D

PR #229, opened 2018-07-05
Backend: Update some deps

jits:
This "fixes" some security vulnerabilities (though not for anything we're directly using / affected by).


jits:
Am looking into the broken build.

PR #230, opened 2018-07-05
Backend: update `sprockets` dep

jits:


jits:
@marcinc - yep, although it's not being used at the moment.

PR #236, opened 2018-07-18
Escape query string when searching for a user

### LINKED ISSUE #233: [bug] 404 redirect immediately when typing user email to add to a project
**Steps to recreate:**
1. Login to the Platform Hub
1. Navigate to a project that you have administrative control over: `https://[HUB-URL]/projects/detail/[PROJECT-NAME]`
1. Start typing a user's email address

**Expected Behaviour:**
- When typing the user's email address, the drop-down will show the related user with matching address

**Actual Behaviour:**
- When typing `.` as part of the email / username, the form automatically submits and immediately redirects you to the homepage. Responses below:

```
Request URI: /api/users/search/bob@?include_deactivated=false
Response Code: 200 OK
Response Message: []

Request URI: /api/users/search/bob.bobberson?include_deactivated=false
Response Code: 404 Not Found
```

KashifSaadat:
Fixes #233

### DIFF
--- a/platform-hub-api/config/routes.rb
+++ b/platform-hub-api/config/routes.rb
@@ -33,7 +33,7 @@ Rails.application.routes.draw do
     end
 
     resources :users, only: [ :index, :show ] do
-      get '/search/:q', to: 'users#search', on: :collection
+      get '/search/:q', to: 'users#search', on: :collection, :constraints => { :q => /[^\/]+/ }
 
       get :identities, on: :member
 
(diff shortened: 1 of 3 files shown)


platform-hub-api/config/routes.rb
```diff
@@ -33,7 +33,7 @@
     end
 
     resources :users, only: [ :index, :show ] do
-      get '/search/:q', to: 'users#search', on: :collection
```
jits: Nice! Would you mind adding a spec case for this in `users_routing_spec.rb`?

KashifSaadat: Sure, added 👍

PR #240, opened 2018-07-27
New help search page (part 1)

jits:
The first stage of many: this PR introduces a new "help search" page where users can search for help in the hub. We currently only index support request templates in ElasticSearch and surface these as the first type of "help items" available. The goal being to index other items, like FAQ entries, docs, etc., and provide a unified help search page.

Note: this is behind a feature flag so can be switched on when it's ready.


platform-hub-api/bin/elasticsearch/README.md
```diff
@@ -0,0 +1,29 @@
...
+
+Scripts to run and interact with a local containerised ElasticSearch instance.
+
+Intended for local dev and not for production environments!
+
+Uses a local data volume to store the data.
+
+See `.env` for details of the set up.
+
+## Prequisites
```
marcinc: Prerequisites

PR #248, opened 2018-09-07
Sync, fetch and index markdown docs from GitHub repos

jits:
This commit introduces the logic / services to:
- sync the markdown file entries from GitHub repos to entries in the db (with support for adding, modifying and deleting entries on subsequent syncs)
- fetching the content of these markdown files, parsing out the headers and overall content, sanitising, and then indexing into the Help Search index
- rake task to trigger the full docs sync whenever needed and on a regular cycle if needed
- UI to configure GitHub repos as sources for docs
- UI to show docs in Help Search results

---

**Note:** there's more to come in separate PRs: GitLab repo support; better admin UI; manual sync option; etc.

### DIFF
--- a/platform-hub-api/Gemfile
+++ b/platform-hub-api/Gemfile
@@ -29,6 +29,7 @@ gem 'api-pagination', '~> 4.7'
 gem 'memoist', '~> 0.16.0'
 gem 'has_scope', '~> 0.7.2'
 gem 'elasticsearch-persistence', '~> 5.1'
+gem 'commonmarker', '~> 0.17.11'
 
 group :development, :test do
   gem 'byebug', platform: :mri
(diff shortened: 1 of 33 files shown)


jits:
@marcinc – on the naming `git_hub` vs `github` – it's unfortunately been a bit inconsistent; _technically_ it should be `git_hub` since that's the underscored version of `GitHub` (which is the official way they spell it). The existing GitHub agent stuff is already using `git_hub`. I think the env vars are the only ones that have it as one word. But moving forward, and for new stuff, it should really be `git_hub` and `GitHub`, I think.

marcinc:
@jits hmm... not a big deal, just doesn't read right.

jits:
@marcinc –

> […] just doesn't read right.

Agreed.

PR #249, opened 2018-09-10
Sync, fetch and index markdown docs from hosted GitLab repos

jits:
- This introduces a new `Agents::GitLabAgentService` that provides access to the hosted GitLab instance
- Both the syncing and doc processing code for GitHub has been abstracted out into a base class that is now used by both the GitHub and GitLab doc sources
- This includes a fix for parsing out the headings in markdown docs where further markup is used in the heading string itself (e.g. to make the heading bold)


platform-hub-api/app/services/agents/git_lab_agent_service.rb
```diff
@@ -0,0 +1,23 @@
+module Agents
+  class GitLabAgentService
+
+    attr_reader :base_url
+
+    def initialize base_url:, token:
+      @base_url = base_url
+      @client = Gitlab.client(
+        endpoint: URI.join(base_url, 'api/v4').to_s,
```
marcinc: Could potentially extract API version to configuration.

PR #273, opened 2018-11-08
Initial work on Docker Repos support

jits:
Note: this is not yet functional:
- It lays out the barebones UI and hub backend for supporting Docker Repo management for projects.
- In upcoming work we'll hook this up to the backend integration service (via a message bus) and add access management.

### DIFF
--- a/platform-hub-api/app/models/ability.rb
+++ b/platform-hub-api/app/models/ability.rb
@@ -53,6 +53,16 @@ class Ability
     end
 
 
+    # Docker repos in projects
+
+    can :read_docker_repos_in_project, Project do |project|
+      can_participate_in_project project, user
+    end
+    can :administer_docker_repos_in_project, Project do |project|
+      can_participate_in_project project, user
+    end
+
+
     # Users
 
     can :identities, User do |target_user|
--- a/platform-hub-api/app/models/docker_repo.rb
+++ b/platform-hub-api/app/models/docker_repo.rb
@@ -1,0 +1,51 @@
+class DockerRepo < ApplicationRecord
+
+  NAME_REGEX = /\A[a-z]+[a-z0-9\-_\/]*\z/
+
+  include Audited
+
+  audited descriptor_field: :name, associated_field: :service
+
+  attr_readonly :name, :service_id
+
+  before_validation :build_repo_name
+
+  belongs_to :service, -> { readonly }
+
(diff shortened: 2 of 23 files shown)


platform-hub-api/app/models/ability.rb
```diff
@@ -53,6 +53,16 @@ def initialize(user)
     end
 
 
+    # Docker repos in projects
+
+    can :read_docker_repos_in_project, Project do |project|
+      can_participate_in_project project, user
+    end
+    can :administer_docker_repos_in_project, Project do |project|
+      can_participate_in_project project, user
```
marcinc: This should be `can_administer_project` ?

platform-hub-api/app/models/docker_repo.rb
```diff
@@ -0,0 +1,51 @@
...
+  before_validation :add_project_shortname_to_name
+
+  belongs_to :service, -> { readonly }
+
+  scope :by_services, ->(ids) { where(service_id: ids) }
+
+  enum status: {
+    pending: 'pending',
+    ready: 'ready',
+    deleting: 'deleting',
```
marcinc: Just a thought. What if repo couldn't get created upstream for whatever reason? Should we add `failed` status (or similar) or we just keep it permanently in `pending` in such case?

platform-hub-api/app/models/docker_repo.rb
```diff
@@ -0,0 +1,51 @@
...
+  def readonly?
+    if persisted?
+      read_only_attrs = self.class.readonly_attributes.to_a
+      if read_only_attrs.any? {|f| send(:"#{f}_changed?")}
+        raise ActiveRecord::ReadOnlyRecord, "#{read_only_attrs.join(', ')} can't be modified"
+      end
+    end
+  end
+
+  def add_project_shortname_to_name
```
marcinc: maybe name it in a more generic way e.g.`build_docker_repo_name` ?

platform-hub-api/app/models/ability.rb
jits: I think the [consensus in #hub_ecr](https://hod-dsp.slack.com/archives/CDMFKN5TJ/p1541601224025900) was that any one in a project can create or delete repos.

platform-hub-api/app/models/docker_repo.rb
marcinc: Cool 👍

PR #275, opened 2018-11-16
Initial flows for viewing and requesting updates to the access policy for Docker repos

jits:
Will be hooked up to the backend integration service in upcoming PR(s).

### DIFF
--- a/platform-hub-api/app/models/docker_repo.rb
+++ b/platform-hub-api/app/models/docker_repo.rb
@@ -2,7 +2,14 @@ class DockerRepo < ApplicationRecord
 
   NAME_REGEX = /\A[a-z]+[a-z0-9\-_\/]*\z/
 
+  ACCESS_STATUS = {
+    pending: 'pending',
+    active: 'active',
+    removing: 'removing'
+  }.freeze
+
   include Audited
+  include ValidateHashes
 
   audited descriptor_field: :name, associated_field: :service
 
@@ -16,7 +23,7 @@ class DockerRepo < ApplicationRecord
 
   enum status: {
     pending: 'pending',
-    ready: 'ready',
+    active: 'active',
     deleting: 'deleting',
   }
 
@@ -34,7 +41,32 @@ class DockerRepo < ApplicationRecord
 
   validates :service_id, presence: true
 
+  validate_hashes(
+    access: {
+      schema: {
+        'robots' => [[
+          {
+            'username' => /\A[a-z]+[a-z0-9\-_]*\z/i,
+            'status' => DockerRepo::ACCESS_STATUS.values.to_set,
+          }
+        ]],
+        'users' => [[
+          {
+            'username' => String,
+            'writable' => TrueClass,
+            'status' => DockerRepo::ACCESS_STATUS.values.to_set,
+          }
+        ]],
+      },
+      unique_checks: [
+        { array_path: 'robots', obj_key: 'username' },
+        { array_path: 'users', obj_key: 'username' },
+      ]
+    }
+  )
+
   attr_default :status, :pending
+  attr_default :access, -> { { 'robots' => [], 'users' => [] } }
 
   private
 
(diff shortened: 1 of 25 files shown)


platform-hub-api/app/models/docker_repo.rb
```diff
@@ -16,7 +23,7 @@ class DockerRepo < ApplicationRecord
 
   enum status: {
     pending: 'pending',
-    ready: 'ready',
+    active: 'active',
     deleting: 'deleting',
```
KashifSaadat: should this now be `removing` instead, as per the `ACCESS_STATUS` defined above?

jits: It's a fair suggestion; my thinking here was that the `DockerRepo` is standalone parent object, which actually does get deleted from the universe, but the individual robot or user within the access policy of the repo just gets removed – i.e. the IAM account may not actually get deleted, so that robot and/or user may not technically be deleted, it's just that they've been removed from this particular repo. Do you think that's enough reason to have this slight difference in wording?

KashifSaadat: Ah I understand, yup I think that makes sense :)

platform-hub-api/app/models/docker_repo.rb
```diff
@@ -2,7 +2,14 @@ class DockerRepo < ApplicationRecord
 
   NAME_REGEX = /\A[a-z]+[a-z0-9\-_\/]*\z/
 
+  ACCESS_STATUS = {
+    pending: 'pending',
+    active: 'active',
+    removing: 'removing'
```
KashifSaadat: When a user has moved from `active` to `removing` and this has been processed, does the record drop altogether (so they don't linger in a `removed` state)?

platform-hub-api/app/models/docker_repo.rb
```diff
@@ -34,7 +41,32 @@ class DockerRepo < ApplicationRecord
 
   validates :service_id, presence: true
 
+  validate_hashes(
+    access: {
+      schema: {
+        'robots' => [[
+          {
+            'username' => /\A[a-z]+[a-z0-9\-_]*\z/i,
```
KashifSaadat: Just raising as a note, this and the repository `NAME_REGEX` is fine for ECR but may be different with other integrations depending on what they support.

platform-hub-api/app/models/docker_repo.rb
jits: :) just to clarify further: in upcoming work I'll be actually deleting stuff when responding to the results from the backend integration service.

platform-hub-api/app/models/docker_repo.rb
jits: Yes, good point. Hopefully these are fairly conservative anyway, and thus forward compatible. But if not, we'll need to scope these checks by type.

platform-hub-web/src/app/projects/projects-detail.html
```diff
@@ -334,8 +334,100 @@ <h4>{{m.user.email}}</h4>
...
+                  </md-expansion-panel-content>
+                </md-expansion-panel-expanded>
+              </md-expansion-panel>
             </md-card-content>
             <md-card-actions layout="row" layout-align="start center">
+              <md-button
+                ng-disabled="d.status != 'active'"
+                aria-label="Manage the access for this Docker repo"
+                ng-click="$ctrl.dockerRepoManageAccess(d, $event)">
+                Manage Access
```
KashifSaadat: Access management doesn't appear to be scoped to project admins, just confirming this is intentional right? So the plan initially is that all project members can create, delete and control access for docker repos.

jits: Yep, this is intentional and a key design decision of this new piece of work (i.e. that project team members have more autonomy over management of resources).

PR #280, opened 2018-11-30
Docker repos: hook up to SQS - send tasks and route results from the backend

jits:


platform-hub-api/app/services/docker_repo_access_policy_service.rb
```diff
@@ -1,5 +1,7 @@
 class DockerRepoAccessPolicyService
 
+  RESOURCE_TYPE = 'DockerRepositoryAccessPolicy'.freeze
```
marcinc: just a question: could this be a symbol instead? similarly below in lifecycle service?

jits: Not easily – elsewhere in the PR we use this to match the `resource_type` from the results message, so it's easier if it's always a string.

PR #289, opened 2019-08-19
Scan images via Anchore

KashifSaadat:
- Added anchore image scan steps for CI
- Simplified the CI pipeline around the API tests, image build and validation
- Use the official ruby-alpine docker image for the api image build
- Update FactoryGirl to FactoryBot and fix deprecation warnings

CI job: https://drone.acp.homeoffice.gov.uk/UKHomeOffice/platform-hub/1406


platform-hub-api/.dockerignore
```diff
@@ -27,5 +27,4 @@ pgdata
 .ruby-version
 .rspec
 .gitignore
-spec
```
jits: Don't need this change anymore?

platform-hub-api/Dockerfile
```diff
@@ -1,15 +1,14 @@
...
     && apk --update add ca-certificates libressl \
     && update-ca-certificates \
-    && apk --update add \
-    bash curl ruby ruby-irb ruby-rake ruby-io-console ruby-json ruby-bigdecimal libstdc++ tzdata \
-    postgresql-client \
+    && apk --update add bash curl postgresql-client postgresql-dev \
+    && apk -U add --virtual build_deps build-base make \
     && echo 'gem: --no-document' > /etc/gemrc \
-    && gem install bundler -v 1.16.2
+    && gem install bundler -v 2.0.2
```
jits: I recall Bundler v2+ having some issues, so may be worth sticking with `1.16.2`? Also, the `Gemfile.lock` has been explicitly built with `1.16.2`, which will only cause a warning when `bundle install` is run, but may as well stick to this version unless there are security patches for it?

platform-hub-api/.dockerignore
KashifSaadat: Good spot, was meant to revert that!

platform-hub-api/Dockerfile
KashifSaadat: There wasn't anything that came up for bundler from what I recall so will revert it back 👍

PR #303, opened 2020-09-09
Projects improvements

Rhodine:
### Issue
Users visiting the Projects page have difficulty finding their projects, as they have to scroll through a long list of projects which includes projects of which they are not members. See ticket [#2025](https://github.com/UKHomeOffice/application-container-platform-board/issues/2025) for more information.
### Solution
I have organised the list of projects on the Projects page so that it prioritises and highlights the user's projects at the top of the list with all other projects greyed out (sorted alphabetically). This includes changes to the projects-list view and the addition of tests. 
### Process
In **projects-list.component.html**: To highlight the user’s projects, I added a ‘Member’ badge if the user is a member of the project and I also greyed out any projects of which the user is not a member:
In **projects-list.component.js**: To Display the User's Projects, I joined the project membership object to the project object and then filtered them based on whether they contained a project member with an id that was the same the current user id, assigning ‘true’ status to the ‘isProjectTeamMember’ property of those that did.
To display Projects of which User was Not a Member, I filtered the results based on whether they contained a project member with an id that was not the same the current user id, assigning ‘false’ status to the ‘isProjectTeamMember’ property of those that did.
To order the projects, I ordered the results for the user's projects and the projects not belonging to the user by name in ascending order separately. I then concatenated them, with results for the user's projects as the first argument.
### Testing - Unit tests
- To test all projects were shown with the user's projects prioritised, I created a mock projects list and tested that the user's projects appeared in the first position and projects of which the user was not a member appeared in the second position of the Projects.all object. I also tested that the html tag `md-card` with and id `projects` existed, which would only be the case if a list of projects existed.
- To test that user projects had ‘Member’ badges attached, I created a mock projects list where the only existing project was the user’s project and tested that the html tag `small` with the id `member-badge` existed.
- To test that projects of which the user was not a member did not have a member badge attached, I created a mock projects list where the only existing project was not a project the user was a member of and tested that the html tag `small` with the id` member-badge `did not exist.

### Issue
This resolves the issue of project admins having to raise a ticket for the request to be resolved by ACP support. See ticket [#1132](https://github.com/UKHomeOffice/application-container-platform-board/issues/1132) for more information.
### Solution
I have added functionality for project admins to make and demote other project admins. This includes changes to the the projects-detail view, ability modeland the projects controller tests.
### Process
In **projects-detail.html**, I have made the ‘Make admin’ and ‘Demote admin’ buttons visible to project admins as well as hub admins.
In **projects-detail.component.js** I have amended the makeAdmin and demoteAdmin functions to include project admins in this exception with the logic.
In **ability.rb**, I have amended the model to assign :set_role and :unset_role capability to project admins(using the private method can_administer_project).

I also updated puma (from 3.10.0 to 3.12.6) and rack (from 2.0.7 to 2.2.3) to pass Anchore scan(see Gemfile.lock).
### Testing - Integration tests
- To test that project admins could make other project admins, I amended the current `PUT #set_role` and `DELETE #unset_role`tests and added tests for the context in which the user is not a hub admin but is project admin of the same project, and the context in which the user is not a hub admin but is project admin of a different project. 

## Impact
The code does not appear to break existing functionalities. Manual testing on the test instance shows new display of projects list and there does not seem to be any performance impacts.  Project admins making and demoting other project admins works as it currently does for hub admins and there doesn't seem to be any delays. Manual testing was carried out on Safari and Google browsers.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -127,7 +127,7 @@ pipeline:
   api_scan_image:
     image: quay.io/ukhomeofficedigital/anchore-submission:latest
     image_name: platform-hub-api:${DRONE_COMMIT_SHA}
-    whitelist: CVE-2019-5419,CVE-2019-5420,CVE-2020-8130,CVE-2020-11077,CVE-2020-11076,CVE-2020-5247,CVE-2019-16770 
+    whitelist: CVE-2019-5419,CVE-2019-5420,CVE-2020-8130,CVE-2020-11077,CVE-2020-11076,CVE-2020-5247,CVE-2019-16770,CVE-2020-8164,CVE-2020-8165,CVE-2020-8162,CVE-2019-5420 
     when:
       event: [push, tag]
 
--- a/platform-hub-api/spec/controllers/projects_controller_spec.rb
+++ b/platform-hub-api/spec/controllers/projects_controller_spec.rb
@@ -638,6 +638,54 @@ RSpec.describe ProjectsController, type: :controller do
 
       end
 
+      context 'not a hub admin but is project admin of same project' do
+        before do
+          create :project_membership_as_admin, project: @project, user: current_user
+        end
+
+        context 'when user is not a project team member' do
+          it 'should return a 400 Bad Request error' do
+            put :set_role, params: { id: @project.id, user_id: @user.id, role: role }
+            expect(response).to have_http_status(400)
+          end
+        end
+
+        context 'when user is a project team member' do
+          before do
+            create :project_membership, project: @project, user: @user
+          end
+
+          it 'should set the role accordingly' do
+            expect(ProjectMembership.exists?(project_id: @project.id, user_id: @user.id, role: role)).to be false
+            expect(Audit.count).to eq 0
+            put :set_role, params: { id: @project.id, user_id: @user.id, role: role }
+            expect(response).to be_success
+            expect(json_response['role']).to eq role
+            expect(ProjectMembership.exists?(project_id: @project.id, user_id: @user.id, role: role)).to be true
+            expect(Audit.count).to eq 1
+            audit = Audit.first
+            expect(audit.action).to eq 'set_role'
+            expect(audit.auditable.id).to eq @project.id
+            expect(audit.associated.id).to eq @user.id
+            expect(audit.user.id).to eq current_user_id
+            expect(audit.data['previous_role']).to eq nil
+            expect(audit.data['new_role']).to eq role
+          end
+        end
+      end
+
+      context 'not a hub admin but is project admin of a different project' do
+        before do
+          another_project = create :project
+          create :project_membership_as_admin, project: another_project, user: current_user
+        end
+
+        it 'should not be able to set the role - returning 403 Forbidden' do
+          put :set_role, params: { id: @project.id, user_id: @user.id, role: role }
+          expect(response).to have_http_status(403)
+        end
+      end
+
     end
   end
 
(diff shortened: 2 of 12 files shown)


.drone.yml
```diff
@@ -127,7 +127,7 @@ pipeline:
   api_scan_image:
     image: quay.io/ukhomeofficedigital/anchore-submission:latest
     image_name: platform-hub-api:${DRONE_COMMIT_SHA}
-    whitelist: CVE-2019-5419,CVE-2019-5420,CVE-2020-8130,CVE-2020-11077,CVE-2020-11076,CVE-2020-5247,CVE-2019-16770 
+    whitelist: CVE-2019-5419,CVE-2019-5420,CVE-2020-8130,CVE-2020-11077,CVE-2020-11076,CVE-2020-5247,CVE-2019-16770,CVE-2020-8164,CVE-2020-8165,CVE-2020-8162,CVE-2019-5420
```
rdabin, line 130: rather than adding exceptions to whitelist here is it possible to resolve any of these? e.g. CVE-2020-8164 looks like using rails>6.0.3.1 fixes this - not sure if this is in scope/simple enough to upgrade major version of rails, but worth discussing.

platform-hub-api/spec/controllers/projects_controller_spec.rb
```diff
@@ -700,6 +700,103 @@ def expect_result result
...
+            expect(audit.auditable.id).to eq @project.id
+            expect(audit.associated.id).to eq @user.id
+            expect(audit.user.id).to eq current_user_id
+            expect(audit.data['previous_role']).to eq role
+            expect(audit.data['new_role']).to eq nil
+          end
+        end
+      end
+
+      context 'not a hub admin but is project admin of a different project' do
```
rdabin, line 788: looks like this context repeated above on L739, can setup there be reused & `it` block be moved there?

platform-hub-web/src/app/projects/projects-list.component.js
```diff
@@ -23,18 +26,66 @@ function ProjectsListController(roleCheckerService, Projects) {
...
+    return Projects.getAll()
+    .then(projects => {
+      return Me.refresh().then(() => {
+        const currentUserId = Me.data.id;
+        return Promise.all(projects.map(projectIdto => {
+          return Projects.getMemberships(projectIdto.id);
+        })).then(memberships => {
+          ctrl.memberships = memberships;
+
+          ctrl.myProjects = getMyProjects(projects, currentUserId, memberships);
```
rdabin, line 38: here i'd probably have one functions `getProjects` and pass a flag like `myProjects: boolean`, that way can reuse most or whats in `getMy/NotMyProjects`

platform-hub-web/src/app/projects/projects-list.component.js
```diff
@@ -23,18 +26,66 @@ function ProjectsListController(roleCheckerService, Projects) {
...
+          ctrl.notMyProjects = getNotMyProjects(projects, currentUserId, memberships);
+
+          ctrl.myProjects = _.orderBy(ctrl.myProjects, [project => project.name.toLowerCase()], ['asc']);
```
rdabin, line 41: may as well move this inside `getProjects` if refactoring to single function as mentioned in above comment^

platform-hub-web/src/app/projects/projects-list.html
```diff
@@ -17,14 +17,16 @@ <h3>
 
   <loading-indicator loading="$ctrl.loading"></loading-indicator>
 
-  <md-content>
+  <md-content id="projects">
 
-    <md-card ng-repeat="p in $ctrl.Projects.all track by p.id">
+    <md-card id="projects" ng-repeat="p in $ctrl.Projects.all track by p.id"
```
rdabin, line 22: has same id as `<md-content>` above - id should be unique or use class

platform-hub-api/spec/controllers/projects_controller_spec.rb
Rhodine: I have resolved this now - test moved to correct describe block.

platform-hub-web/src/app/projects/projects-list.html
Rhodine: Resolved - extra id removed from `<md-content>`

.drone.yml
Rhodine: This is addressed in another ticket as it is a task in itself.

platform-hub-api/spec/controllers/projects_controller_spec.rb
```diff
@@ -700,6 +700,103 @@ def expect_result result
 
       end
 
+      context 'not a hub admin but is project admin of same project' do
+        before do
+          create :project_membership_as_admin, project: @project, user: current_user
+        end
+
+        context 'when user is not a project team member' do
+          it 'should return a 400 Bad Request error' do
```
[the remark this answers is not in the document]
Rhodine: This was existing implementation. No breaking changes but worth considering another MR to refactor.

platform-hub-web/src/app/projects/projects-list.component.js
Rhodine: Resolved- created one function `getProjects`

platform-hub-web/src/app/projects/projects-list.component.js
Rhodine: Resolved - moved into refactored getProjects function

rdabin (approved):

